PR #7, opened 2020-05-24
Fix chanced output bug in multis

hjae78:


src/main/java/gregicadditions/machines/multi/simple/LargeSimpleRecipeMapMultiblockController.java
```diff
@@ -159,7 +159,7 @@ protected Recipe createRecipe(long maxVoltage, IItemHandlerModifiable inputs, IM
 
 		protected void copyChancedItemOutputs(RecipeBuilder<?> newRecipe, Recipe oldRecipe, int multiplier) {
 			for (Recipe.ChanceEntry s : oldRecipe.getChancedOutputs()) {
-				int chance = s.getChance() * this.chancePercentage / 100;
+				int chance = Math.min(10000, s.getChance() * this.chancePercentage / 100);
 				int boost = s.getBoostPerTier() * this.chancePercentage / 100;
```
huneau, line 163: should do the same for boost ? what do you think ?

PR #113, opened 2020-07-17
Various bug fixes, config option to disable all ore variants in void miner

hjae78:


src/main/java/gregicadditions/recipes/RecipeHandler.java
```diff
@@ -285,42 +285,52 @@ private static void processWireGt(OrePrefix wireGt, IngotMaterial material) {
                 case 0:
                 case 1:
                     RecipeMaps.ASSEMBLER_RECIPES.recipeBuilder().circuitMeta(24 + cableSize).input(OrePrefix.wireGtSingle, material, cableAmount).input(foil, Rubber, cableAmount).outputs(cableStack).duration(150).EUt(8).buildAndRegister();
+                    break;
```
huneau, line 288: Here the non break is expected. I would like that lower tier recipe can be made with higher tier recipe

src/main/java/gregicadditions/recipes/RecipeHandler.java
```diff
@@ -285,42 +285,52 @@ private static void processWireGt(OrePrefix wireGt, IngotMaterial material) {
...
                     RecipeMaps.ASSEMBLER_RECIPES.recipeBuilder().circuitMeta(24 + cableSize).input(OrePrefix.wireGtSingle, material, cableAmount).input(foil, PolyphenyleneSulfide, cableAmount).outputs(cableStack).duration(150).EUt(8).buildAndRegister();
-
+                    break;
             }
         }
         switch (tier) {
             case 0:
             case 1:
                 RecipeMaps.ASSEMBLER_RECIPES.recipeBuilder().circuitMeta(24).input(wireGt, material).input(foil, Rubber, cableAmount).outputs(cableStack).duration(150).EUt(8).buildAndRegister();
+                break;
```
huneau, line 314: Here is the same comment. Non break it's expected

hjae78:
Ah that make sense, will push

huneau:
I finally merge the PR I let you close all related issue.

PR #125, opened 2020-08-07
Update the lang file of Chinese Simplified

magcicada:
Without strict inspection, there may be mistakes, you can merge first, and then we will check for mistakes later.


huneau:
thx @magcicada

PR #253, opened 2020-11-24
Update ru_ru.lang file

KRP-sudo:
Merged update of localization file


huneau:
huge work, thx so much :)

PR #263, opened 2020-12-05
Vanadium, Barium and Uranium

pablocnvs:


huneau:
Its a good job, but Its not finished, there is missing translation  you can add your translation here => https://github.com/Gregicality/gregicality/blob/master/src/main/resources/assets/gtadditions/lang/en_us.lang

And then, Uranium Oxide and Thorium Oxide has no usage.
You have to hook it with the actual nuclear process. if you need help just tell me

huneau:
@pablocnvs

pablocnvs:
should be done by now

pablocnvs:
actually, don't merge it yet, since i'm going to add recipes to make it possible to get metallic uranium and thorium from the chain

huneau:
nice job, its better like this. There is some missing part but I can take care, if you want to do it I can explain to you in private

PR #265, opened 2020-12-08
Add RecipeMap cast to CTLargeRecipeMap

brachy84:
Why? Because removeByOutput needs a RecipeMap. With this cast it works fine.


huneau:
ah really interesting!
nice job

PR #274, opened 2020-12-12
Lithium Titanate

pablocnvs:
I made a small pr to add the recipe for lithium titanate, which i'll be using, but leaving you the fusion components


huneau:
@pablocnvs can you go back to our current 0.21 stuff? otherwise, I have to fix your PR again, and its quite painful to not make a mistake

PR #282, opened 2020-12-21
Supporting GT tools in AE2

Yefancy:
This is a very simple coremod, almost no hacking of AE2. 
Feature: fixed an issue with NBT meta items (such as gt tools) that could not be recognized, when using JEI to import a recipe into the crafting terminal.

ae2

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -28,6 +28,13 @@ minecraft {
     replaceIn("Gregicality.java")
 }
 
+jar {
+    manifest {
+        attributes 'FMLCorePluginContainsFMLMod': 'true'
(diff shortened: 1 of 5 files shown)


huneau:
houuuuuuuuuuaaaaaaaahhhhhhh sooooo nice !!!!!!!!

huneau:
@hjae78 what do you think ?

hjae78:
This is incredible! @Yefancy is it possible to fix this for refined storage as well or does this already do that?

Yefancy:
So far it only works with AE2. i play RS so rarely that i forgot it and ill fix this for RS later.

PR #284, opened 2020-12-22
Supporting metaItem imported AE2+RS

Yefancy:
The content is basically the same as #282 .
additional fixes:

- importing from player's inventory
- Refined Storage

ae2+rs


huneau:
really impressive

PR #293, opened 2020-12-30
Prospecting Tool GUI [TODO]

Yefancy:
- [x] Widget ProspectingMap
- [x] Widget OreOilList
- [x] Better Render
- [x] Logic
- [x] ~~Sound~~
- [x] Progressive


Yefancy:
It's not perfect, but I think can merge the basic gui frame now.

huneau:
pffffff really nice ! you are awesome  !

huneau:
what do you mean about logic ? 
the last thing I want its the underground oil only valid for LuV and ZPM because we have the solar sampler

Yefancy:
> what do you mean about logic ?
> the last thing I want its the underground oil only valid for LuV and ZPM because we have the solar sampler

logic almost done for me. maybe some details need to be modified according to your idea.
i might add a tooltip to the map when mouseover a chunk, prompting the amount of ore / oil.

Yefancy:
^ commit 2d5c79909ab307aaae5d5339d8d8aed828f663ad
this should be my last functional update. i also tested it in client / server.
do you have any changes that you can commit directly, and I won't commit anything of prospector other than fixing bugs.

PR #322, opened 2021-01-17
Fix helium compatibility problem with NuclearCraft

LasmGratel:
**Issue**
With both GTCE and NuclearCraft installed, `liquid_helium` fluid is registered by GTCE but with different temperature
By default, GTCE Helium fluid has 300K of temperature while NC has only 4K.

According to https://github.com/turbodiesel4598/NuclearCraft/blob/master/src/main/java/nc/fluid/SuperFluid.java#L8 and https://github.com/turbodiesel4598/NuclearCraft/blob/6abd59c1d3dec75db6d60a439d36d91bad039ec9/src/main/java/nc/init/NCFluids.java#L49

Thus, [QMD](https://github.com/Lach01298/QMD) will be unable to cool itself down normally
image

**Workaround**
change to fluid material and set the temperature
(plus idk how to increase ID so please interview it)

**Temporal CraftTweaker solution**:
```javascript
#loader gregtech
import mods.gregtech.material.MaterialRegistry;
MaterialRegistry.createFluidMaterial(7891, "liquid_helium", 0xDDDD00, "gas").setFluidTemperature(4).addFlags(["STATE_GAS", "GENERATE_PLASMA"]);
```


LasmGratel:
Alternatively, you can add a `setFluidTemperature` method in `SimpleFluidMaterial` so we don't have to convert it into heavy FluidMaterial class every time

hjae78:
Thanks for the PR and bringing this to my attention, however I don't want to add new normal materials at this time. I will add a way to configure heat of SimpleFluidMaterial and expose it to CraftTweaker.

PR #323, opened 2021-01-17
[Fix] prospector underground oil crashing

Yefancy:
[crash-report](https://discord.com/channels/701354865217110096/704380911206006855/800042679769104434)

The original method was redundant and inapplicable:

- Use id instead of biomeName;
- Fix and retain the method to get the biomeName;


hjae78 (approved):

hjae78:
Thanks!

PR #326, opened 2021-01-17
Large Packaging Machine

TechLord22:
Adds a large packaging machine, which can use the Packaging and Unpackaging RecipeMaps. The multiblock is gated behind LuV, as the casing material and controller requires Naquadah EBF Coils.


hjae78 (approved):

hjae78:
Great work!

PR #330, opened 2021-01-22
Stop non-fusion related multiblocks from using superconducting and fusion coils

TechLord22:
This prevents the superconducting and fusion coils from working in the EBF, Volcanus, Chemical Plant, Pyrolyse Oven, and Cracker Unit. They are also removed from the machines' JEI previews. 

There is only one potential recipe issue - Abyssal Alloy. This material uses 9625K in the EBF. This change would prevent it from being created until the Advanced Heating Coil Mk1. This coil requires Trinium, which requires the Mk2 Void Miner, which requires many materials using UHV voltage. Nearly every UHV singleblock, transformer, and energy hatch uses Abyssal Alloy cables or wires - making progressing to UHV impossible. The solution to this is to change the coil to not utilize Trinium in its recipe. Though, I believe this may have already been addressed and fixed elsewhere. 

Additionally, I do not think it is possible to remove the EBF and Multismelter info tooltips from the Superconducting and Fusion Coils, so hiding/removing them and replacing them with Gregicality's own versions may be necessary.


huneau:
hum interresting 🤔

hjae78:
I had this exact idea but never got around to it! Please add a config option for this please

YoungGT:
Config needed.

hjae78:
As regards to Abyssal Alloy I'd rather just see its temperature lowered

src/main/java/gregicadditions/GAConfig.java
```diff
@@ -1123,7 +1124,23 @@
 
         }
 
+        public static class HeatingCoils {
 
+            @Config.Comment("GregTech CE heating coils to blacklist from working in non-fusion multiblock machines. Default=\"superconductor\", \"fusion_coil\"")
+            @Config.Name("GTCEHeatingCoilBlacklist")
```
hjae78, line 1161: Please include spaces in the name parameter

hjae78 (changes requested):

hjae78 (approved):

hjae78:
Good job!

PR #334, opened 2021-01-24
Assorted small fixes

TechLord22:
Fixes 3 things:
1) Fixes the missing UIV Arc Furnace by adding the missing registry entry. There was an ID gap between the UEV and UMV arc furnaces
2) Removes duplicate Acetone and Ethenone recipes in the fluid heater
3) Removes the old Neuro Processing Unit recipe from the assembly line


hjae78:
Thank you!

PR #335, opened 2021-01-26
Cosmic Circuits

pablocnvs:


src/main/java/gregicadditions/recipes/chain/CosmicComponents.java
```diff
@@ -0,0 +1,341 @@
...
+                .fluidOutputs(RadonRadiumMix.getFluid(288))
+                .buildAndRegister();
+
+        ADV_FUSION_RECIPES.recipeBuilder().duration(100).EUt(900000).coilTier(3).euStart(10000000000L)
+                .fluidInputs(ScandiumTitanium50Mix.getFluid(144))
+                .fluidInputs(RadonRadiumMix.getFluid(144))
+                .fluidOutputs(MetastableHassium.getFluid(144))
+                .buildAndRegister();
+
+        FUSION_RECIPES.recipeBuilder().fluidInputs(Americium.getFluid(144), Titanium.getFluid(144)).fluidOutputs(Tennessine.getFluid(288)).duration(100).EUt(75000).EUToStart(520000000).buildAndRegister();
```
hjae78, line 65: There should be Advanced Fusion recipes

hjae78 (changes requested):

src/main/java/gregicadditions/recipes/chain/CosmicComponents.java
pablocnvs: I thought it was going to be too much, but then i remembered that people use AE or RS to automate it, so it will probs be tier 2 advanced fusion

a-thoren (approved):

hjae78:
Good job!

PR #337, opened 2021-01-27
Make regular fusion reactors use the matching casing and coil tier

TechLord22:
Modified the Fusion Reactors Mk. 1-3 to use Mk 1, Mk 2, and Mk 3 casings and coils respectively. Also updated the JEI previews to reflect this. Changed the recipes for the fusion casings 2 and 3 to reflect the new tearing. Mk 2 uses Rutherfordium, and Mk 3 uses Dubnium. Fusion coil recipes were updated as well, using LuV, ZPM, and UV power in the Assembly Line for the appropriate tier.


hjae78 (approved):

hjae78:
Nice job! Want to get decals opinion on this as well

huneau:
The first look seems ok for me, 
the second look, this will break the compatibilité of a known multiblock...

hjae78:
It will indeed break existing multiblocks but we have done that several times before, overall I think it is a necessary change.

huneau (approved):

huneau:
ok for me then

PR #339, opened 2021-01-29
Steam Grinder Multiblock

DStrand1:
The Steam Grinder:
SteamGrinderOverview

The Steam Grinder (adapted from GT++) is a Steam-Era multiblock to serve as an alternative to the terribly slow Steam, LV, and even MV Macerators. It's only downside past that is that it is not very efficient at doing single crafts (similar to the Multismelter), and has no secondary outputs. It can process up to 8 of an item at once, using formulas:

`duration = Math.max(recipeDuration, recipeDuration * 1.5)`
`totalEUt  = Math.min(32, Math.ceil(recipeEUt * 1.33))`

This simplifies roughly to `duration * 1.5` and `recipeEUt * 1.33`, but handles some edge cases.

Note that this multiblock uses new Steam I/O busses, and a new Steam Hatch, which is basically an Energy Input Hatch, but for Steam. It will NOT form with normal I/O busses, and does not accept normal Fluid I/O hatches either.
DoesntForm

Here are the recipes:
GrinderRecipe
SteamHatchRecipe
SteamInputBusRecipe
SteamOutputBusRecipe

And the tooltip for the Steam Grinder block:
ControllerTooltip

Note that the secondary metal used here is Tumbaga, matching how it was done in GT++. Because I wanted to preserve this, I changed Tumbaga to be a smeltable dust instead of an EBF dust. It still requires some Gold to make, but this way, it gives players a reason to push for Gold early, and try to get to the better path of acquiring it. All in all, this multiblock does not require too much Tumbaga, so it is still achievable in Steam/LV.

Additionally, I added the GT++ Style shapeless recipes for Tumbaga, by adding a new dust, Tumbaga Mix. If this change seems unnecessary, and we want to keep this machine gated by a Mixer, I can revert it.

Lastly, the JEI preview:
MultiblockPattern

In order to create this, I had to create many parent classes capable of handling a multiblock that uses Steam. This means that we can relatively easily add new Steam Multiblocks if we wish to in the future. Note that the parent class `RecipeMapSteamMultiblockController` is coded specifically to NOT allow fluid I/O other than the Steam Hatch, which is not included in recipe matching, so this will limit the abilities of Steam multiblocks.

I am confident in my coding of the parent classes and recipe logic, but I am not certain that the material (and Tile Entity) registration was done perfectly. I would like at least a close look on those sections of this PR, to make sure I do not accidentally introduce an ID shift.

Thank you for considering this addition to Gregicality, as it took a lot of work to build these parent classes, and to create the recipe logic allowing for multiple items of a single recipe to process at once.


src/main/java/gregicadditions/GAMaterials.java
```diff
@@ -1310,7 +1310,7 @@
     public static Material UXV = new MarkerMaterial("UXV");
     public static Material MAX = new MarkerMaterial("MAX");
 
-    //free id 775
+    public static DustMaterial TumbagaMix = new DustMaterial(775, "tumbaga_mix", 0xFFB20F, MaterialIconSet.DULL, 2, of(new MaterialStack(Gold, 2), new MaterialStack(Bronze, 1)), 0);
```
hjae78, line 1313: I don't understand why this is added? You say that it's to introduce a shapeless recipe (?) for Tumbaga but then why not just add a recipe using the exisiting Tumbaga material?

src/main/java/gregicadditions/machines/GATileEntities.java
```diff
@@ -1118,6 +1127,12 @@ public static void init() {
         LARGE_PACKAGER = GregTechAPI.registerMetaTileEntity(4172, new TileEntityLargePackager(location("large_packager"), RecipeMaps.PACKER_RECIPES));
         GregTechAPI.registerMetaTileEntity(4173, new TileEntityLargePackager(location("large_packager"), RecipeMaps.UNPACKER_RECIPES));
         COSMIC_RAY_DETECTOR = GregTechAPI.registerMetaTileEntity(4174, new MetaTileEntityCosmicRayDetector(location("cosmic_ray_detector")));
+
+        STEAM_HATCH = GregTechAPI.registerMetaTileEntity(4200, new MetaTileEntitySteamHatch(location("steam_hatch")));
```
hjae78, line 1131: This won't cause any ID shifts as it simply registers the tile entity at the number you specify, you absolutely don't need to skip forward in the ID here and I'd prefer if we continued where we previously left at 4174.

src/main/java/gregicadditions/machines/multi/impl/GAMultiblockAbility.java
```diff
@@ -0,0 +1,12 @@
+package gregicadditions.machines.multi.impl;
+
+import gregtech.api.metatileentity.multiblock.MultiblockAbility;
+import net.minecraftforge.fluids.IFluidTank;
+import net.minecraftforge.items.IItemHandlerModifiable;
+
+public class GAMultiblockAbility<T> extends MultiblockAbility<T> {
+
+    public static final MultiblockAbility<IFluidTank> STEAM = new MultiblockAbility();
+    public static final MultiblockAbility<IItemHandlerModifiable> STEAM_IMPORT_ITEMS = new MultiblockAbility<>();
```
hjae78, line 10: Are you sure this needs to be in a new class the extends MultiblockAbility? We have new MultiblockAbility declared in GregicAdditionsCapabilities, where the Qubit hatches are registered.

src/main/java/gregicadditions/machines/multi/impl/RecipeMapSteamMultiblockController.java
```diff
@@ -0,0 +1,128 @@
...
+
+    @Override
+    protected boolean checkStructureComponents(List<IMultiblockPart> parts, Map<MultiblockAbility<Object>, List<Object>> abilities) {
+        //basically check minimal requirements for inputs count
+        int itemInputsCount = abilities.getOrDefault(GAMultiblockAbility.STEAM_IMPORT_ITEMS, Collections.emptyList())
+                .stream().map(it -> (IItemHandler) it).mapToInt(IItemHandler::getSlots).sum();
+        return itemInputsCount >= recipeMap.getMinInputs() &&
+                abilities.containsKey(GAMultiblockAbility.STEAM);
+    }
+}
```
hjae78, line 128: How come you can't extend RecipeMapMultiblockController? As far as I can see all you're doing differently is changing recipemapWorkable

src/main/java/gregicadditions/recipes/GARecipeAddition.java
```diff
@@ -535,6 +535,20 @@ public static void init() {
...
+        MIXER_RECIPES.recipeBuilder()
+                .duration(200)
+                .EUt(30)
+                .input(dust, Gold, 7)
+                .input(dust, Bronze, 3)
+                .outputs(OreDictUnifier.get(dust, Tumbaga, 10)).buildAndRegister();
+        ModHandler.addShapelessRecipe("tumbage_from_mix", OreDictUnifier.get(dust, Tumbaga, 10),
+                OreDictUnifier.get(dust, TumbagaMix), OreDictUnifier.get(dust, TumbagaMix), OreDictUnifier.get(dust, TumbagaMix),
+                OreDictUnifier.get(dust, Gold));
+        ModHandler.addShapelessRecipe("tumbaga_mix_shapeless", OreDictUnifier.get(dust, TumbagaMix),
```
hjae78, line 548: Ah now I see why you added Tumbaga Mix. It should be SimpleDustMaterial though. Also these recipes can have ItemStacks with size larger than 1 I'm quite sure

src/main/java/gregicadditions/GAMaterials.java
DStrand1: Tumbaga is 7 Gold dust and 3 Bronze dust in a mixer, so there wasn't any way I could make it handcraftable otherwise

src/main/java/gregicadditions/machines/GATileEntities.java
DStrand1: Alright, I will adjust it. I left some space cause when I was working on this, the Cosmic Ray TE hadn't been added yet, as it was in a very recent commit

src/main/java/gregicadditions/machines/multi/impl/GAMultiblockAbility.java
DStrand1: Ok, I will refactor to this. I didn't see that we had a place for it already.

src/main/java/gregicadditions/machines/multi/impl/RecipeMapSteamMultiblockController.java
DStrand1: I wanted to remove the notion of Fluid inputs to make Steam easier to implement. In hindsight, what you are suggesting may have been possible though. If you'd like me to restructure to that, I can

src/main/java/gregicadditions/recipes/GARecipeAddition.java
DStrand1: I tried that, and it showed itemstacks with amounts of greater than one in the shapeless recipe, instead of breaking them up. Then when crafting, it only took one of that material. Kind of a weird system, but whatever

huneau:
I not sure to want that, but The only thing I can say its thanks  lot for your job and its seams really well done. I will make you a review later

hjae78:
Even if decal declines this I think you've done a great job and this opens up a lot of options for the future and you should definitely release this as a standalone addon!

src/main/java/gregicadditions/GAMaterials.java
huneau: it should be removed

src/main/java/gregicadditions/machines/multi/impl/RecipeMapSteamMultiblockController.java
```diff
@@ -0,0 +1,128 @@
+package gregicadditions.machines.multi.impl;
```
huneau, line 1: it should be placed in gregicadditions.capabilities.impl

src/main/java/gregicadditions/machines/multi/impl/SteamMultiblockRecipeLogic.java
```diff
@@ -0,0 +1,122 @@
+package gregicadditions.machines.multi.impl;
```
huneau, line 1: it should be placed in gregicadditions.capabilities.impl

src/main/java/gregicadditions/machines/multi/steam/MetaTileEntitySteamGrinder.java
```diff
@@ -0,0 +1,219 @@
...
+                currentRecipe = previousRecipe;
+            }
+
+            if (currentRecipe != null && setupAndConsumeRecipeInputs(currentRecipe)) {
+                setupRecipe(currentRecipe);
+            }
+        }
+
+        @Override
+        protected Recipe findRecipe(long maxVoltage, IItemHandlerModifiable inputs, IMultipleTankHandler fluidInputs) {
```
huneau, line 119: I trust you on algo

src/main/java/gregicadditions/recipes/GARecipeAddition.java
huneau: should be removed, because tumbaga is for advanced stuff

huneau (changes requested):

huneau:
if @DStrand1 fix all comment @hjae78 you can merge

DStrand1:
With regards to extending `RecipeMapMultiblockController` instead of `MultiblockWithDisplayBase` for my parent class, I have to do this for 3 reasons.
1. To remove the notion of fluid inputs other than Steam
2. To allow for a different front Overlay texture
3. To properly use my new RecipeLogic class, as otherwise I would not be able to use `SteamMultiblockRecipeLogic`, since I had to have it extend `AbstractRecipeLogic` and not `MultiblockRecipeLogic`.

All other requested changes have been resolved and committed.

hjae78:
Okay it looks good now, but I can see an issue with the recipe since as decal said Tumbaga is for higher tier stuff. The EBF temperature should be restored. Instead we could use Potin, this does not have an EBF temperature and is alloyed from early game materials

DStrand1:
Alright, Tumbaga is back to where it is on master. I changed all my Steam Grinder components that used Tumbaga to use Potin instead, similar to the Large Macerator. As a bonus, it is not gated by anything, so it solves my problems and yours too. This should be ready to merge now.

hjae78:
Great! Thanks for your hard work!

PR #343, opened 2021-01-31
Fix cupriavidus necator culture ID and internal name

TechLord22:
Resolves ID conflict created with the cosmic circuits. Also fixes misspelled internal name.


hjae78:
Thank you for this, but Zalgo opened a PR where he added textures and addressed the ID issue however the misspelt internal name I will address in the future

PR #344, opened 2021-01-31
Extend Large Circuit Assembly Line

TechLord22:
The large circuit assembly line now has it's own RecipeMap, supporting Qubits, with 4 fluid inputs and 16 item inputs. The regular circuit assembler no longer supports Qubits. 

In order to reflect the RecipeMap changes for the large circuit assembly line, the structure can be extended with slices in the same way as the Assembly Line. 

All the circuits in the regular circuit assembler have had their recipes added to the Large Circuit Assembly Line. The recipes which used higher voltage than the circuit assembler supported have been removed from it. Additionally, the circuits that were previously made in the Assembly Line have moved to the Circuit Assembly Line. The "Processing Units" such as the Cyber Processing Unit remain in the Assembly Line.


hjae78 (approved):

hjae78:
I like it, want to get decals opinion before accepting!

huneau:
its huge work but I dont understand the purpose of this PR

hjae78:
It's so we can move all circuit recipes into the large circuit assembler from Assembly Line. If we have Circuit Assembly Line why do we make them in normal Assembly Line?

huneau:
I dont understand because I dont see diff between large circuit assembler and assembly line. 
With that PR we have 2 machines, made in the same ways, with the same resources but with split recipe maps.
Actually, the project is not clear from my side. 

I can understand but it needs more work to provide new experiences with the large circuit assembler.
You can probably look impact stuff to get cool ideas or imagine something really awesome, Im aware with that.
My expectation is that :
- the current large simple circuit assembler have to stay here (because its currently well balanced otherwise we have to balance it again)
- you can add a new tier of large machines for UHV+ circuits.
- you can remove the qubit stuff from circuit assembler and add it to a new tier of recipe map

I hope Im not frustrating you to much

hjae78:
Hmm, you make good points. Mainly that then the Large Circuit Assembler and Assembly Line will be pretty much the same machine with similar resources. I'm not sure what changes could be made to differentiate it from the Assembly Line. Perhaps it is best to leave the Large Circuit Assembler as it is and introduce a new multiblock for higher tier wafers?

pablocnvs:
Since we'll need a machine for the wafers anyways, making a bit different and integrating the wafers into the large circuit assembler (maybe with a rename) is a logical step.

TechLord22:
Closing for now until a better solution is implemented in the Wafer Overhaul.

PR #346, opened 2021-01-31
Various Changes from #suggestions

TechLord22:
This PR mostly includes things from the suggestions channel on discord. 

1. Allows placing of functional blocks on the top layer of the chemical plant. There wasn't any good reason to disallow this, so this has been re-enabled to enhance automation capabilities. 
2. Fix the output of nonmetal waste to use buckets for fluids. The new values are the bucket equivalents of the ingot values.
3. Changes the advanced fusion reactor to use one advanced fusion coil in its recipe, instead of four. This prevents the player from having 3 leftover coils after building the reactor with Mk 1 coils.
4. Fixed localization for a few things. The Plate Field Shape now mentions "Plate." Also, Alkaline Earth Sulfate now uses the spelling of Sulfur/Sulfate used for everything else.


hjae78:
👍

PR #350, opened 2021-02-04
Transformer TOP Update

DStrand1:
This PR updates our TOP info for transformers to be consistent with the new information displayed in GTCE. I felt it was redundant to add new localization keys since they already existed in GTCE's lang files, but this meant that the key would only resolve if the GTCE version is 1.11.1 or greater.

If GTCE 1.11.1+ is loaded:
gtce1 11 1

If a lower version is loaded (1.8 as is in my dev environment):
gtce1 8 x

If you would like me to change this to not require GTCE 1.11.1+, it would be an easy fix, but would just mean redundant lang entries. Personally, I think it is okay that we do not show this information if GTCE does not either.

### DIFF
--- a/src/main/java/gregicadditions/theoneprobe/GATransformerProvider.java
+++ b/src/main/java/gregicadditions/theoneprobe/GATransformerProvider.java
@@ -1,0 +1,67 @@
+package gregicadditions.theoneprobe;
+
+import gregicadditions.GAUtility;
+import gregicadditions.GAValues;
+import gregicadditions.machines.energy.GAMetaTileEntityTransformer;
+import gregtech.api.capability.IEnergyContainer;
+import gregtech.api.metatileentity.MetaTileEntity;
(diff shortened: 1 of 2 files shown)


a-thoren (approved):

hjae78 (approved):

huneau:
hoooooooouuuuu fancy ! we should dont care about version, we are up to date or nothing.

Let me check, there is a default TOP fix to do to avoid weird render like `error: gtaddition are noobs` something like that

huneau:
na ok its good for me 

but @hjae78 keep in mind, when there is default capability for oneprobe there should be list there https://github.com/Gregicality/gregicality/blob/8a8b36a833843981020513b60dc2f1c8a95f48d4/src/main/java/gregicadditions/capabilities/SimpleCapabilityManager.java

huneau (approved):

PR #353, opened 2021-02-04
Fix wafer localizaiton

TechLord22:
Updates the localization to `[Name] Wafer` instead of `Wafer [Name]` for the following items:
1. ARAM Wafer
2. UHPIC Wafer
3. UHASoC Wafer
4. HASoC Wafer
5. Optical SoC Wafer


hjae78:
👍

PR #354, opened 2021-02-04
Cable Diodes

DStrand1:
Past versions of GT, specifically most recently GT5U with Bartworks, added blocks similar to Transformers, except instead of changing the voltage, they change the amperage. The Cable Diode allows 1-way energy flow, as well as the ability to step cables down to lower amperages.

In this example, I have an MV transformer stepping power down to LV, meaning I have a total of 4A on the LV line. With the cable diode, the 1x cable past it does not burn, as only 1A is being passed through:
diode4to1x

However, this machine allows more than just 1A max through if you want it to. By hitting it with a Soft Hammer, you can toggle its amperage mode, which will cycle through [1, 2, 4, 8, 16] looping around.
diodechangemode

Now, if we put that 1x cable down, it will burn, since 2A are being passed through:
diodeburnsin2a
However, if we had put a 2x cable down, it would not burn.

I followed convention similar to the transformers for the recipes, giving a recipe for each tier following this pattern:
diodeRecipeTooltip

And I added them for every tier, excluding ULV and MAX:
alldiodesandTOP
If you look closely, I even added some TOP info similar to transformers to avoid confusion :)

A few primary use cases for these:
- If you have a 16x line of cables, full 16A, and have a machine 20 blocks away you want to power, you can use a Cable Diode to be able to use only 1x cables instead of 16x for that 20 block distance.
- You can use them as 1-way power passthroughs. For example, if I have 4A being passed through a Diode, on the output side of it, I can add more power to the line, without worrying about the input side of the Diode being over-amped. This way, you can again save on cables, using lower thicknesses instead of making the entire line a large thickness.
- And of course, the obvious use, you can simply step down cables now, without ever needing to worry about burning.


huneau:
yeah nice

huneau (approved):
good for me

hjae78 (approved):

hjae78:
👍  Thank you!

PR #356, opened 2021-02-06
Steam Oven

DStrand1:
**What:**
This PR expands on the newly added Steam Multiblock parent classes, and adds a new multiblock inspired by the Railcraft Steam Oven, the brother to the Steam Grinder in 1.7.10 GT packs.

**How solved:**
The Steam Oven:
Overview

It follows convention with the boilers, and lights up the Firebox Casings when active:
active

I chose to use Firebox Casings for this since I wanted to make it a bit cheaper than the Steam Grinder, as Railcraft's was a 2x2x2 multiblock.

The recipe, which is also a little cheaper than the Steam Grinder, but not much (uses Invar for gears, like the MultiSmelter):
recipe

And lastly, the tooltip and JEI preview:
tooltip
JEIPreview

**Outcome:**
Added Steam Oven

**Additional info:**
- I separated the nested `SteamGrinderWorkable` class and added a parameter for the number of items to process. I left it at 8 for this multiblock as well, but this makes it more flexible for the future. 
- I also added a new config entry for the `Steam -> EU` conversion rate for Steam Multiblocks. The standard rate used by single block Steam machines is 1:1, but for these, I decided to set their default to 2 Steam -> 1EU to balance it. It is still saving Steam overall compared to single block machines though.

### DIFF
--- a/src/main/java/gregicadditions/GAConfig.java
+++ b/src/main/java/gregicadditions/GAConfig.java
@@ -460,6 +460,7 @@ public class GAConfig {
         public AdvFusion advFusion = new AdvFusion();
         public LargeEngraver largeEngraver = new LargeEngraver();
         public HeatingCoils heatingCoils = new HeatingCoils();
+        public SteamMultis steamMultis = new SteamMultis();
     }
 
         public static class LargeEngraver {
@@ -1384,4 +1385,10 @@ public class GAConfig {
                 GALog.logger.info("Added resevoir type " + fluid);
             }
         }
+
+    public static class SteamMultis {
+        @Config.Comment({"Steam to EU multiplier for steam multiblocks. 1.0 means 1 Steam -> 1EU. 2.0 means 1 Steam -> 2EU. 0.5 means 2 Steam -> 1EU"})
+        @Config.RequiresMcRestart
+        public static double steamToEU = 0.5;
     }
+}
--- a/src/main/java/gregicadditions/jei/multi/simple/SteamOvenInfo.java
+++ b/src/main/java/gregicadditions/jei/multi/simple/SteamOvenInfo.java
@@ -3,6 +3,7 @@ package gregicadditions.jei.multi.simple;
 import com.google.common.collect.Lists;
 import gregicadditions.machines.GATileEntities;
 import gregtech.api.metatileentity.multiblock.MultiblockControllerBase;
+import gregtech.common.blocks.BlockFireboxCasing;
 import gregtech.common.blocks.BlockMetalCasing;
 import gregtech.common.blocks.MetaBlocks;
 import gregtech.integration.jei.multiblock.MultiblockInfoPage;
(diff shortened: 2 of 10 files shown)


huneau:
should fix the multiblock and even the grinding with the correct ga metal casing

huneau:
otherwise its a good job :)

src/main/java/gregicadditions/jei/multi/simple/SteamOvenInfo.java
```diff
@@ -0,0 +1,49 @@
...
+    @Override
+    public List<MultiblockShapeInfo> getMatchingShapes() {
+        ArrayList<MultiblockShapeInfo> shapeInfo = new ArrayList<>();
+
+        shapeInfo.add(MultiblockShapeInfo.builder()
+                .aisle("FFF", "IXX", "###")
+                .aisle("HFF", "S#X", "XXX")
+                .aisle("FFF", "OXX", "###")
+                .where('S', GATileEntities.STEAM_OVEN, EnumFacing.WEST)
+                .where('X', MetaBlocks.METAL_CASING.getState(BlockMetalCasing.MetalCasingType.BRONZE_BRICKS))
```
hjae78, line 34: As decal said, change this line to use the Gregicality Bronze casing as we use that as a standard

hjae78 (changes requested):

src/main/java/gregicadditions/jei/multi/simple/SteamOvenInfo.java
DStrand1: I tried changing it to `GAMetaBlocks.getMetalCasingBlockState(Materials.Bronze)` and it threw a NPE. Am I doing this wrong?

src/main/java/gregicadditions/GAConfig.java
```diff
@@ -1384,4 +1385,10 @@ else if (index == 1) {
                 GALog.logger.info("Added resevoir type " + fluid);
             }
         }
+
+    public static class SteamMultis {
+        @Config.Comment({"Steam to EU multiplier for steam multiblocks. 1.0 means 1 Steam -> 1EU. 2.0 means 1 Steam -> 2EU. 0.5 means 2 Steam -> 1EU"})
```
aimiebell, line 1390: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Method names should start with a lower case letter (NM_METHOD_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-method-names-should-start-with-a-lower-case-letter-nm-method-naming-convention).
Methods should be verbs, in mixed case with the first letter lowercase, with the first letter of each internal word capitalized.

hjae78: What, that's a class not a method

hjae78:
Are we currently stuck on the firebox casings being mismatched with the generated bronze casing?

DStrand1:
Yes, this PR is still held up by issues with the casings:

- Generated casings would have the wrong recipe, using Frames instead of Bricks
- Generated casings would have a different color from Firebox Casings
- There is problems with generating a new Firebox, not sure if it is possible
- These same problems would now be present in the Large Bronze Boiler

hjae78:
@huneau it is ok for me to proceed with the casing as is. We can imagine these Bronze casings as a special type of casing like we have for other machines.

huneau:
merge it, I will fix it

PR #357, opened 2021-02-06
Textures for supracausal stuff

Zalgo239:
-Added wormhole generator textures
-Added neutronium sphere textures
-Added Supracausal SMDs textures


hjae78:
I really like them but I feel the neutronium spheres might need some changes. Are they 32x32?

Zalgo239:
> 
> 
> I really like them but I feel the neutronium spheres might need some changes. Are they 32x32?

They are 64x64 actually, I can probably remake those this evening to be 32x32 and a little more interesting

hjae78:
I'd prefer if we could have them be 16x16, they would be consistent with most spherical shapes we have then

Zalgo239:
Alright, I'll make them 16x16 then

hjae78:
Looks great! Thank you so much!

PR #358, opened 2021-02-06
Many Recipe Changes

TechLord22:
This PR makes many changes, many of which improving QOL significantly.

1. Puts Blizz powder's recipe in terms of ingots.
2. Blaze powder can be made similarly to Blizz powder.
3. Ender pearl dust can be created from it's components.
4. Removes a duplicate Quartz Block decompression crafting recipe created by GTCE.
5. Unified Lithium Carbonate Solution to reduce confusion and streamline optical circuit crafting. There was no reason for a regular and a "pure" version.
6. An alternative recipe for Boric Acid allows many late game processes to not be bottlenecked by Brine Processing.
7. Laser Cooling Unit no longer uses UHV circuits and now uses LuV. This prevents UV Optical Processors from needing UHV circuits (it doesn't make sense for earlier circuits to require later ones), as well as offsets the high cost a bit more.
8. Indium Phosphide now uses UHV power, allowing Optical Circuits to be made at around UHV-UEV instead of much later.
9. Zinc Sulfide also now uses UHV power, for the same reasons as Indium Phosphide
10. Scandium Titanium-50 had incorrect English localization. This is now fixed.
11. Fixed all Rare Earth Element Oxide formulas. They all now follow the format: `Element2O3`. Also changed the EBF recipes to go from Oxide -> Regular to fit the new formulas. This is technically a nerf, so the outputs of the REE process may need to be buffed.
12. An alternate Strontium Sulfate recipe allows for Halide Lamps to be produced without a convoluted process.
13. An alternate Ammonium Vanadate recipe was added for 2 reasons. The first is for Vanadium (V) Oxide, making Halide Lamps easier for the same reason as above. The second is to make Pure Sodium Vanadate, which is used for Optical Circuits, making this process more streamlined as well. This added one new SimpleDustMaterial: Sodium Metavanadate, which prevents infinite byproduct loops in proceses. It is also the technically correct version of Sodium Vanadate, but they should not be unified for the reason previously stated.
14. Added an alternate recipe for Tantalum Oxide according to the following equations: 2 Ta + 5 Cl2 → 2 TaCl5 and 2 TaCl5 + 5 H2O → Ta2O5 + 10 HCl. They have been merged into one recipe to avoid another SimpleMaterial: 2 Ta + 5 Cl2 + 5 H2O -> Ta2O5 + 10HCl.
13. Added many recipes to go from Rare Earth Element -> Oxide, making later components not require unprocessed materials from the REE process. All future recipes should follow this format: `2Element + H2SO4 -> Element2O3 + Diluted Sulfuric Acid`.
14. An alternate recipe for Tungsten Trioxide (does not interfere with tungsten process) was added to make Cadmium Tungstate less of a problem when making Scintillators for Optical Circuits.
15. Added an alternate Magnesium Chloride recipe for use in producing Acetylating Reagent, also used to eventually produce Scintillators.
16. Added some missing EBF temperatures for some recipes.
17. Fixed the recipe for producing Lu/Tm:YVO Precipitate.
18. Fix some circuit recipes by changing the fluids used to be in terms of ingots.
19. Fixed Optical and Cosmic Resistors outputting 1 item instead of 32.

Changes Regarding Supra-Causal Circuits:
1. Change the Time Dilation Unit to be significantly cheaper. 
2. Changed circuit Processing Cores to not use cores of the previous circuit tiers.
3. Changed Wormhole recipes to be significantly easier.

The Supra-Causal Circuits need to have many more changes outside of this PR. They are currently far too expensive.


hjae78 (changes requested):

src/main/java/gregicadditions/recipes/GARecipeAddition.java
```diff
@@ -537,6 +547,13 @@ public static void init() {
                 .fluidOutputs(SupercooledCryotheum.getFluid(10000))
                 .buildAndRegister();
 
+        MIXER_RECIPES.recipeBuilder().EUt(60).duration(880)
```
hjae78, line 550: I think this recipe might be better off at HV

hjae78 (changes requested):

huneau (approved):

hjae78 (approved):

hjae78:
Good job! Thanks for your help with this

PR #360, opened 2021-02-08
Supra-Causal Circuit Rebalance

TechLord22:
This PR rebalances and fixes the Supra-Causal circuits. Previously, the Supra-Causal Mainframe used around 16 million platinum foil, among other things - and was simply too difficult to be realistically made. The aim of this PR is to drastically reduce the cost of the circuits and make them worth their very high cost. 

Comparing the supra-causal processor to the cosmic assembly, it isn't completely cheaper, but it is much easier to make overall, and I believe it will be worth making for the player. The supra-causal assembly is far cheaper than the cosmic computer. The supra-causal computer uses more tiered components and circuits than the cosmic mainframe, but uses less complicated materials overall. The supra-causal mainframe is still very expensive, but is actually possible by this point of the game.

The emitters, sensors, and field generator recipes for LuV-UEV were changed to be significantly cheaper. The items are no longer apart of the end-game tiers, and do not need to be as expensive as they used to be. Another significant change made for nearly everything, but especially these, was to reduce the Soldering Alloy usage. It previously reached around 155,000 buckets for the mainframe, and now sits at around 300. Next, the yield for many high use materials was increased. These were primarily Oganesson, Hassium, Quark Degenerate Matter, Rhenium Degenerate Matter, and Neutronium.

One important set of changes were those done to the wormholes. They were previously far to expensive to ever be made, and have since become far cheaper. Additionally, the yield for Neutronium spheres being increased to 4 is extremely important, else the Mendevelium, Naquadria, and Taranium usage becomes entirely unmanageable. All four charges were made much cheaper as well, as more than 200 are required for the supra-causal processor alone. Another big change is the reduction in superconductor usage. Previously, nearly every recipe using superconductors used 32 or 64 wires, which was far too much, thus making the cost in pumps far higher than it needed to be. 

Some components were changed from UMV to UIV and UHV. This is because UMV components would use UEV circuits in their recipes, which would mean using UEV causal circuits to make more UEV causal circuits. The same goes for UEV and higher voltage field generators, since they use the circuit of their voltage tier. This is clearly not good, so they have been lowered accordingly. 

Other changes are smaller, but do still make an impact large enough to be worth using. 

The recipe costs were calculated by modifying patterns in a Refined Storage network. I do not recommend attempting this yourself with the original recipes, especially for the mainframe, on account of using approximately 125,000 ZPM emitters and more. I manually increased the timeout limit significantly in the RS config, and while it didn't take too long to calculate each time, it is still quite intensive. Please ask if you would like me to calculate a more detailed cost breakdown for specific parts of the circuit recipes or overall.


huneau (approved):

src/main/java/gregicadditions/recipes/chain/UHVMaterials.java
```diff
@@ -71,12 +70,12 @@ public static void init() {
...
         ASSEMBLY_LINE_RECIPES.recipeBuilder()
-                .input(plateDense, Mendelevium.getMaterial())
-                .inputs(GELLED_TOLUENE.getStackForm(16))
-                .input(block, Naquadria, 8)
-                .input(block, Taranium)
-                .inputs(DEGENERATE_RHENIUM_PLATE.getStackForm(4))
-                .fluidInputs(Tritanium.getFluid(1296))
+                .input(plate, Mendelevium.getMaterial())
+                .inputs(GELLED_TOLUENE.getStackForm(4))
+                .input(ingot, Naquadria)
```
hjae78, line 75: I don't like ingots in recipes like this, would prefer a plate, screws or bolts

hjae78 (commented):
There are also some conflicts with your other PR which I just merged

src/main/java/gregicadditions/recipes/chain/UHVMaterials.java
hjae78: Alternatively we could make a new dust made of Naquadria and Taranium and use 2 of that

hjae78 (approved):

hjae78:
Great job! Really appreciated

PR #367, opened 2021-02-10
Fix Naquadah Chain Chemical Reactions

TechLord22:
This PR fixes the naquadah chain chemical reactions in order to fit the new GTCE standards, as well as fix the incorrect reactions in general. All three naquadah types have zero leftover byproducts, and everything is recycled throughout the process as a result of these changes. Most changes are minor formula modifications and outputs. The most "major" change is the Potassium Metabisulfate, which has been replaced with Potassium Peroxymonosulfate - which is a little harder to make, but shouldn't be an issue at LuV. Huge thanks to @DStrand1 for help with figuring out chemical reactions and balancing equations.


huneau:
you break the mol rules

hjae78:
> you break the mol rules

In the newest version GTCE overhauled how they handled mols, instead of 2 Hydrogen + 1 Oxygen = 3 Water its not 2 Hydrogen + 1 Oxygen = 1 Water. I believe these recipes follow that rule too

TechLord22:
@hjae78 is correct. This process now follows the new mole rules and thus looks different. It does still result in one naquadah compound dust producing one regular dust, so there isn't a buff to naquadah output. But, you no longer loose materials for every naquadah produced which makes the entire process a closed loop. Eventually this will end up being done for every process and this is just the first step.

hjae78:
@huneau for me this is okay, GTCE changed their chemical mol rule and we should follow. We have massive inconsistencies that need to be addressed but this is a good start. As I understand it the amount of recipes you need to complete should be exactly the same, it's just the amount of inputs and outputs being changed which shouldn't change anything

DStrand1:
The Chemistry Changes to GTCE:
https://github.com/GregTechCE/GregTech/pull/1414

I authored these changes, and can vouch for this PR being consistent with what GTCE now follows.

hjae78:
@huneau are we okay to merge this?

PR #368, opened 2021-02-10
More Rhenium Options

pablocnvs:
With this PR Rhenium becomes 'easier' to obtain, since there's currently only one source of it and it is heavily used in UHV+. The tradeoff is that the way to extract it requires some setup and takes advantage of the Bio-Reactor, which isn't much used outside the Wetware/Bioware line.


hjae78 (approved):

hjae78:
Looks good!

PR #373, opened 2021-02-22
Fix Some Outstanding Issues

### LINKED ISSUE #168: [Bug] Weird Mixer recipes for Potassium Metabisulfite and Potassium Disulfate
For some reason these take 2 different Integrated Circuits.

As far as I understand the value of the Circuit is supposed to be the amount of different Ingredients.

### LINKED ISSUE #271: [BUG] Snow <> Snowball 4 instead of 9
The current compressor/macerator/forge hammer/large forge hammer requires/gives 9 snowballs for a snow block, which contradicts the vanilla and GTCE recipes of 4.

### LINKED ISSUE #315: Bending Cylinder, Small Bending Cylinder do not inherit material colors
image

Using latest version of GTCE/Gregicality, the Bending cylinders do not appear to properly inherit the colors of their components, the way that ingots, plates, and other tool parts and tools do.

### LINKED ISSUE #366: On the formula of aluminum in blast furnace
image
image
image
image
In fact, it just work on the first recipe

DStrand1:
- Fixes #315:
See https://github.com/Shadows-of-Fire/Shadows-of-Greg/pull/106 for a description of the fix for this issue.

- Fixes #168:
This issue occurred for any Mixer recipe that had an integrated circuit, but there only happen to be two so far. I added a section of code in the Large Mixer recipe builder to iterate over all inputs, and if it finds a circuit, it removes it from the input list. I also had to save that circuit value, and use it instead of the default logic of `total input items/fluids count` to avoid a conflict.

- Fixes #366:
I added a circuit of configuration 1 to the standard Aluminium EBF recipe. The EUt, duration, and blastTemp are all identical.

- Fixes #271:
I added the flag `EXCLUDE_BLOCK_CRAFTING_RECIPES` and added each recipe manually to use 4 instead of 9.

Each issue I fixed has a different commit, so if you want to see the individual changes for each issue, you can view that commit's changes.

### DIFF
--- a/src/main/java/gregicadditions/recipes/RecipeHandler.java
+++ b/src/main/java/gregicadditions/recipes/RecipeHandler.java
@@ -462,16 +462,30 @@ public class RecipeHandler {
     }
 
     public static void registerLargeMixerRecipes() {
-        RecipeMaps.MIXER_RECIPES.getRecipeList().forEach(recipe ->
-                GARecipeMaps.LARGE_MIXER_RECIPES.recipeBuilder()
-                        .notConsumable(new IntCircuitIngredient(recipe.getInputs().size() + recipe.getFluidInputs().size()))
-                        .EUt(recipe.getEUt())
-                        .duration(recipe.getDuration())
-                        .fluidInputs(recipe.getFluidInputs())
-                        .inputsIngredients(recipe.getInputs())
-                        .outputs(recipe.getOutputs())
-                        .fluidOutputs(recipe.getFluidOutputs())
-                        .buildAndRegister());
+        RecipeMaps.MIXER_RECIPES.getRecipeList().forEach(recipe -> {
+            List<CountableIngredient> inputList = new ArrayList<>();
+            IntCircuitIngredient circuitIngredient = null;
+
+            for (CountableIngredient input : recipe.getInputs()) {
+                if (!(input.getIngredient() instanceof IntCircuitIngredient))
+                    inputList.add(input);
+                else
+                    circuitIngredient = (IntCircuitIngredient) input.getIngredient();
+            }
+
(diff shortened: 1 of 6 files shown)


src/main/java/gregicadditions/recipes/RecipeHandler.java
```diff
@@ -462,16 +462,30 @@ public static void registerLaserEngraverRecipes() {
...
+            IntCircuitIngredient circuitIngredient = null;
+
+            for (CountableIngredient input : recipe.getInputs()) {
+                if (!(input.getIngredient() instanceof IntCircuitIngredient))
+                    inputList.add(input);
+                else
+                    circuitIngredient = (IntCircuitIngredient) input.getIngredient();
+            }
+
+            if (circuitIngredient == null)
```
hjae78, line 476: I think this can open potential conflicts in the future. Imagine a recipe that takes a circuit in normal mixer and there exists a subset of the inputs of size equal to the circuit number that has a separate mixer recipe. Then if we run this recipe in Large Mixer we have a conflict. To solve this you should just filter out the circuit and not save it on row 473.

hjae78 (changes requested):

hjae78:
Thanks for the help, really appreciate it!

src/main/java/gregicadditions/recipes/RecipeHandler.java
DStrand1: I added this because if I hadn't, the two recipes I fixed in this change would conflict since they have the same exact inputs, but different amounts. What I did may not be the best solution, but simply using the existing way of setting a circuit value does not work either.

hjae78 (changes requested):

PR #377, opened 2021-02-25
Generify Tuple

DStrand1:
The `Tuple` class (used currently just for a couple Large Multis) is a useful Object to have, and others can potentially use it to help write more organized code. All this PR does is make it generic, add an `equals()` override, and move it to `/util`


huneau:
the multiblock still work ?

DStrand1:
Yes, the behavior in all places where Tuple was used is identical.

huneau (approved):

PR #381, opened 2021-02-26
Add the Disassembler

DStrand1:
This PR adds the **Disassembler**, a new machine which is used to disassemble machines into their base parts.

## Behavior
The Disassembler will essentially reverse the crafting recipe (crafting table only) of a machine by taking it out of `ForgeRegistries.RECIPES` and reversing it. There will not be a recipe built in the Disassembler if a machine has multiple Crafting Table recipes, but it does ignore GTCE -> Gregicality multiblock controller conversion recipes. Types of machines that can be disassembled:
- MetaTileEntity instanceof IEnergyContainer
This captures any single block machine
- MetaTileEntity instanceof MultiblockControllerBase
This captures any multiblock controller
- MetaTileEntity instanceof SteamMetaTileEntity
This captures any Steam machine
- NOT instanceof MetaTileEntityHull (and GAMetaTileEntityHull)
This excludes machine hulls from Disassembly, as I feel like if this is to have a disassembly recipe, it should be in the Arc Furnace.

With these parameters, basically any machine and multiblock controller can be disassembled. The EU/t for the disassembly recipe will be the maximum voltage of the tier of machine being disassembled (128 for an MV machine, for example). The duration is `5s * outputItems`, but I feel there may be better logic that can be done here.

In order to not cause progression issues, if there is a circuit present in the Crafting Table recipe for the machine, it is transformed into the "worst" circuit of that tier so that it can still be used, but is not able to be upgraded further. Typically this is a Mainframe, but LV, MV, and HV are not, but are still the worst circuit of their tier.

I wanted to maintain the ingredient order for the outputs, so in order to do this, I filled empty spaces in the former recipes with Tiny Piles of Ash. This will not fill a completely empty row, as Crafting Table recipes are stored as `Ingredient[][]`, so if it only had two rows of input Ingredients, it will ignore the third. Any tools used in these recipes (like the Bender) are also converted to Ash.

I also added a Config parameter for enabling these recipes. I did this primarily because this adds another recipe to remove when changing recipes for machines with CraftTweaker. I am currently looking into how to add new CraftTweaker methods so that registering and unregistering Disassembly Recipes is easier for modpack authors.

I made the recipe for this machine quite expensive, since it is a very powerful machine to have. It currently uses on-tier circuits to craft, but I am considering changing these to be one tier up, similar to the Circuit Assembler. Any feedback on this would be very appreciated.

## Gallery
image

image

image


src/main/java/gregicadditions/machines/GATileEntities.java
```diff
@@ -1155,6 +1161,18 @@ public static void init() {
             DIODES.add(GregTechAPI.registerMetaTileEntity(id++, new GAMetaTileEntityDiode(location("diode." + GAValues.VN[i].toLowerCase()), i)));
         }
         STEAM_OVEN = GregTechAPI.registerMetaTileEntity(4197, new MetaTileEntitySteamOven(location("steam_oven")));
+
```
hjae78, line 1164: Let's wrap this with the config option as well

src/main/java/gregicadditions/CommonProxy.java
```diff
@@ -195,6 +195,7 @@ public static void registerOrePrefix(RegistryEvent.Register<IRecipe> event) {
 
     @SubscribeEvent(priority = EventPriority.LOWEST)
     public static void registerRecipesLowest(RegistryEvent.Register<IRecipe> event) {
+        DisassemblyHandler.buildDisassemblerRecipes();
```
hjae78, line 198: Moving the config check to wrap this rather than wrapping the entire method is cleaner

src/main/java/gregicadditions/recipes/DisassemblyHandler.java
```diff
@@ -0,0 +1,259 @@
...
+
+                // One of the dumbest blocks of code I've ever written
+                // At least we have proper EUt for multiblock disassembly now
+                for (ItemStack itemStack : outputItems) {
+                    Tuple<ItemStack, Integer> tempTuple;
+                    MetaTileEntity tempMTE;
+
+                    // Check if itemStack is a circuit, and set the voltage to its tier
+                    if ((voltage = ((tempTuple = circuitToUse.get(itemStack.getItemDamage())) != null ? tempTuple.getSecond() : 0)) != 0) {
+                        voltage = 8 << (voltage * 2);
```
hjae78, line 187: Think you could just reference GAValues.V here

hjae78:
Very nice job!

DStrand1:
All requested changes have been implemented.

huneau:
https://github.com/Nukepowered/NukepoweredUtils => made something better
disassembler should be punitive

DStrand1:
I can change it to chanced outputs if you like. They seem to be using 50% chance for each item, +7.5% per tier.

DStrand1:
It now is chanced outputs, being a 50% base chance plus 7.5% per tier. This does mean that disassembling very high tier machines are unlikely to get all of your items back, since there is not much room to overclock them, but if you think that is okay, then this is ready to merge.

DStrand1:
Tech suggested a Config value for chanced outputs, so I added this, with its default being "true", meaning chanced outputs.

huneau (approved):

hjae78 (approved):

PR #382, opened 2021-02-26
Adding Insulation Wire Assembly to cables

pablocnvs:
Simple PR, just to make wire assembly cables useful. If you want to tweak the ratios, feel free to suggest.


huneau (approved):
ok why not

PR #385, opened 2021-02-28
Yet Another PA Change

DStrand1:
First off, this PR includes the commits of #380, so if it or this PR is merged, it will be included in the code base.

This PR optimizes the PA heavily by creating a set of parent classes to allow for a slot in the multiblock controller.
image

This slot is dynamic, as I have it as part of a parent class, `RMapMultiblockWithSlotController`. As per my current implementation, you pass in a class, the slot can hold an item that extends that class. For the PA, I chose `ITieredMetaTileEntity`. With these parent classes, we could optionally use this for circuits in recipes in our Large multiblocks. 

Now, for the reasoning. Consider this example:
- I have a PA with 3 MAX input buses on it
- I have all the slots full of items, none of which are a valid machine to use for recipes

Now our PA will be lagging an amazing amount as it loops through all slots in all input buses, looking for a machine that doesn't exist. And to make this even better, it is comparing against a List of all registered MTE's, for all filled slots in the buses. With this change, it will never check the input buses unless a valid machine is in the controller slot, thus optimizing the PA by a drastic margin.

Additionally, I added some text to its GUI to show the RecipeMap and Parallel Recipes the PA is able to handle given its machine input. This max value also responds to the Config limit.
Capture

Note: This PR will have a merge conflict with #381, since they both add a file `GAGuiTextures`, but I made sure this file was created for both PR's in the same location so it should not be too difficult to resolve.


DStrand1:
This git hierarchy looks very strange, but it is ready to merge now, and all works as intended.

src/main/java/gregicadditions/capabilities/impl/RMapMultiblockWithSlotController.java
```diff
@@ -0,0 +1,95 @@
...
+import net.minecraft.block.Block;
+import net.minecraft.entity.player.EntityPlayer;
+import net.minecraft.item.ItemStack;
+import net.minecraft.nbt.NBTTagCompound;
+import net.minecraft.util.ResourceLocation;
+import net.minecraftforge.items.ItemStackHandler;
+
+import javax.annotation.Nonnull;
+
+public abstract class RMapMultiblockWithSlotController extends GARecipeMapMultiblockController {
```
huneau, line 20: Write class in full word, this help when searching

src/main/java/gregicadditions/gui/GAGuiTextures.java
```diff
@@ -0,0 +1,9 @@
+package gregicadditions.gui;
+
+import gregtech.api.gui.resources.AdoptableTextureArea;
+import gregtech.api.gui.resources.TextureArea;
+
+public class GAGuiTextures {
```
huneau, line 6: you can you the client handler class

huneau (changes requested):
As always, nice job

DStrand1:
Requested changes have been addressed. I also fixed the compilation issue I described in https://discord.com/channels/701354865217110096/792756247689822239/815830216520892436 since it kept me from testing my changes here.

huneau (approved):

hjae78 (approved):

PR #386, opened 2021-02-28
Fix Some Conflicting Recipes

### LINKED ISSUE #384: Some Conflicted Recipes
### Mod Used
Gregicality-1.12.2-0.21.3
gregtech-1.12.2-1.11.1.632-sources
gregtech-1.12.2-1.11.1.632-energy-api
gregtech-1.12.2-1.11.1.632
CodeChickenLib-1.12.2-3.2.3.358-universal

### Issue Found

Barium Oxide with Aluminum dust to Barium Ingot and Barium Aluminate. 
In this situation the 'aluminum dust to Aluminum Ingot' will take precedence over this recipe, which makes this recipe to produce Barium ingot impossible. You may need a circuit to distinguish these two recipes.

The recipe to make Drilling Mud.
The issue exists in this: Bentonite Clay Solution is impossible to make according to my JEI.
And to mix Clay, Bentonite and Water you get Barium Sulfate Solution, which doesn't make sense.
I believe this is a bug.

### My Consideration
In the first issue, no doubt that Barium plays a vital role in game progress. If this recipe doesn't work, it will be rather difficult to go on. For example, Yttrium Barium Cuprate.
And the Second issue, without Drilling Mud it equals to ban the Drilling Rig, which needs Drilling Mud to operate. Well, it doesn't  affect the game progress as critical as above, but it still useful when we start mass production.

DStrand1:
Issue #384 highlighted a total of 4 recipe conflicts that we have.

- Aluminium Dust EBF recipes
Fixed by #373

- Drilling Mud
Fixed at some point in 0.21.4

- LV-IV Sensor/Emitter Assembler Recipes
To fix these, I changed the sensor inputs slightly. 3 of them used flawless gems, which I changed to normal gems for Sensors. The other two used an Ender Pearl/Eye of Ender, which I changed to their dust form for Sensors.
Alternatively, a circuit could be added to all 5 Sensor recipes, but I wanted to avoid another circuit if possible. If you do not like the change I implemented, I can do that (or another suggestion) instead.

- Hypochlorous + NaOH overlap
Changed these to both require circuits, with configurations 1 and 2

Closes #384


hjae78:
Good job!

PR #387, opened 2021-03-01
Fix Issues with Large Rocket Engine

DStrand1:
Previously, the Large Rocket Engine had many, many problems, among them being:
- Not consuming CO2, ever
- Not consuming fuel in some cases, and getting stuck generating power
- Being able to produce massive amounts of power, near max int value

With all these issues combined, you could get the Large Rocket Engine to a state where it was not consuming any inputs, and was permanently outputting nearly a full Amp of MAX power. Now, this is no longer the case.

Additionally, I set a hard limit on production, capping it at 2A UV. It will consume and void any fuel over this limit, further reinforcing the idea that you are meant to limit its flowrate to a specified, chosen rate.

Lastly, I cleaned up its UI, as it had a lot going on and was very hard to read, and inconsistent with the fuel in the tank. Now, it works perfectly, and is much more readable.
image
If fuel is loaded, the color for that text is changed to green, to signal that it is ready to run once secondary inputs are specified.


huneau:
Nice fix!
But, I don't want to cap the output of the large rocket engine.
Now we have more tiered power we can apply more debuff on higher tiers.

DStrand1:
Power limit is removed, I have not yet added a power debuff on higher tiers, but it is currently only 10% efficiency above UV so it may be okay as-is.

huneau (approved):

huneau:
would be interesting if we have a graph with EU output vs fluid consumption

hjae78:
Thoughts about adding a config option to disable efficiency loss? Currently we have an option to disable efficiency on single block but not one for this

DStrand1:
I can do that, or I can incorporate its efficiency calculations into the same singleblock config value and rename it to show that it does both. Whichever you think is the more desirable solution.

PR #394, opened 2021-03-07
Cleanup MachineRecipeRemoval

ALongStringOfNumbers:
This PR cleans up MachineRecipeRemoval and fixes a couple issues with removals in the file.

First off, a config option is added to enable logging of recipe removals. This is useful for checking which removals are failing and which are succeeding. This is set to default false, because normal players should not need to worry about this option and it could slow down loading of the game.

Also in the config file, an option was removed to make Fine Wire recipes from Foils, as this does not exist anymore in GTCE.

**Recipe Removal File**
- I removed Lead from the list of soldering Materials, as GTCE does not use Lead for any Soldering recipes, and this list is only used in this file, for recipes that the Lead does not apply to.
- I removed the Rotor recipe removals, as this recipe does not exist anymore in GTCE
- I removed a bunch of outdated recipes for making GTCE circuits
- Removed a recipe removal that had Epichlorhydrin as an input as this recipe was removed in the GTCE chemistry update
- Removed the easy rocket fuel recipe removal as this recipe was removed in the GTCE chemistry update
- Removed a bunch of recipes that no longer exist in GTCE
- Removed the Iridium Fusion recipe removal, as I also removed its addition in the recipe addition file. This recipe was added and removed by Gregicality, and does not exist in GTCE
- Fixed the Data Orb recipe removal so that it would actually get removed
- Fixed the Nano CPU wafer recipes so that it would actually get removed
- Removed a section removing LuV and ZPM wetware circuits, as this is now done by the loop removing all circuit recipes
- Adjusted some Field Generator recipe removals to match the oredicts used in base GTCE

### DIFF
--- a/src/main/java/gregicadditions/recipes/GAMachineRecipeRemoval.java
+++ b/src/main/java/gregicadditions/recipes/GAMachineRecipeRemoval.java
@@ -1,6 +1,7 @@
 package gregicadditions.recipes;
 
 import gregicadditions.GAConfig;
+import gregicadditions.utils.GALog;
 import gregtech.api.recipes.*;
 import gregtech.api.recipes.ingredients.IntCircuitIngredient;
 import gregtech.api.unification.OreDictUnifier;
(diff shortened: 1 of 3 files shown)


src/main/java/gregicadditions/recipes/GAMachineRecipeRemoval.java
```diff
@@ -280,40 +201,77 @@ public static void init() {
...
+        List<String> names = new ArrayList<>();
+        for (ItemStack s : itemInputs) {
             inputs.add(s);
-        map.removeRecipe(map.findRecipe(Long.MAX_VALUE, inputs, Collections.emptyList(), Integer.MAX_VALUE));
+            if(GAConfig.Misc.enableRecipeRemovalLogging) {
+                names.add(s.getDisplayName() + " x " + s.getCount());
+            }
+        }
+
+        if(map.removeRecipe(map.findRecipe(Long.MAX_VALUE, inputs, Collections.emptyList(), Integer.MAX_VALUE)) && GAConfig.Misc.enableRecipeRemovalLogging) {
```
hjae78, line 218: You could just wrap this with one check of the config

hjae78 (approved):

hjae78:
Thank you, this is highly appreciated!

DStrand1 (approved):

huneau (approved):

PR #398, opened 2021-03-09
Add Fluid Tooltips

DStrand1:
Following the PR to GTCE https://github.com/GregTechCE/GregTech/pull/1519, I updated our code to allow for fluid tooltips for Simple Fluids as well.

There are 2 parts to this PR:
- Fluid tooltips on Minecraft Buckets
These had to be handled separately, so I handled them in our `ClientProxy` class.

- Fluid tooltips in JEI and in machines
These will not be implemented with this PR. I know, weird right? Well let me explain. This PR includes a coremod for a method in GTCE, `formulaHook`, which I added in my GTCE PR for their fluid tooltips. Now, since this PR is not yet released, our tooltips will similarly not be there in these cases either. However, as soon as that feature is included in GTCE, it will be included for us.

This coremod does **NOT** depend on the feature being added to GTCE, or on having the proper version. If that method does not exist to inject code into, it will simply not happen and no issues will arise. In this case, tooltips will only occur for buckets, as is the behavior if you were to test this feature with the latest GTCE release.

Some images of a fluid I added a temporary chemical formula to:
image
image
image

The only reason I am able to get these images is by building a jar of Gregicality with this change, and of GTCE with the PR included.

Included in this PR is the addition of two things to `SimpleFluidMaterial`:
- We are now storing the fluids as a Map of <String, SimpleFluidMaterial> so they are easily and efficiently indexable.
- I added constructors and logic for generating the chemical formula of tooltips, in a way that all of our old constructors will work without issue.

### DIFF
--- a/src/main/java/gregicadditions/ClientProxy.java
+++ b/src/main/java/gregicadditions/ClientProxy.java
@@ -1,5 +1,6 @@
 package gregicadditions;
 
+import codechicken.lib.util.ItemNBTUtils;
 import com.mojang.realmsclient.gui.ChatFormatting;
 import gregicadditions.blocks.GABlockOre;
 import gregicadditions.blocks.GAMetalCasing;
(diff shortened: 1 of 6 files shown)


TechLord22 (approved):
Looks good from what I can tell.

huneau (approved):
yeeeeaaaaaahhhhhh 👍

hjae78 (approved):

huneau:
@DStrand1 can you fix the conflict, I will merge it

DStrand1:
Done

PR #399, opened 2021-03-09
Refactor SimpleMaterials

DStrand1:
In this PR, I changed our Simple Material classes (Dust, Fluid) to be based on a shared parent class, `SimpleMaterial`. The main motivation behind this is because fluid tooltips are a thing now, so this means less repeated code.

Additionally, I added a mechanism for creating a `chemicalFormula` tooltip from a String. This will open up a lot of possibilities with how we can format our chemical formulas. Additionally, I added an Enum that contains common ions and functional groups, to make it a lot easier to format our formulas, `EnumIonsGroups`. The getter methods in this class are named in a C-like style, because I wanted to make sure the calls to them were small in terms of code length, since our `GAMaterials` code lines are already very long.


src/main/java/gregicadditions/materials/EnumIonsGroups.java
```diff
@@ -0,0 +1,99 @@
...
+        return FORMULA;
+    }
+
+    public String fGroup() {
+        return "(" + FORMULA + ")";
+    }
+
+    public String fGroup(int count) {
+        return "(" + FORMULA + ")" + count;
+    }
```
hjae78, line 98: Can we refactor these methods to be more legible?

src/main/java/gregicadditions/materials/EnumIonsGroups.java
```diff
@@ -0,0 +1,99 @@
+package gregicadditions.materials;
+
+public enum EnumIonsGroups {
```
hjae78, line 3: This could be in GAEnums maybe?

hjae78 (approved):

PR #404, opened 2021-03-13
Prevent JEI RecipeMap Overlaps

TechLord22:
Currently, the Large Mixer and Chemical Plant RecipeMaps overlap with the recipe info in JEI, as shown below. This PR changes these two RecipeMaps to be handled similarly to the Assembly Line, thus preventing the overlapping information. For future use, the logic used in the two new classes can be duplicated for any RecipeMap needing more spacing in JEI, where the Chemical Plant applies to regular RecipeMaps, and the Large Mixer applies to LargeRecipeMaps. 

Chemical Plant Before:
image

Chemical Plant After:
image

Large Mixer Before:
image

Large Mixer After:
image

### DIFF
--- a/src/main/java/gregicadditions/recipes/GARecipeMaps.java
+++ b/src/main/java/gregicadditions/recipes/GARecipeMaps.java
@@ -4,8 +4,7 @@ import crafttweaker.annotations.ZenRegister;
 import gregicadditions.gui.GAGuiTextures;
 import gregicadditions.integrations.exnihilocreatio.SieveRecipeMap;
 import gregicadditions.recipes.map.LargeRecipeBuilder;
-import gregicadditions.recipes.map.NuclearReactorBuilder;
-import gregicadditions.recipes.map.RecipeMapAssemblyLine;
+import gregicadditions.recipes.map.*;
 import gregicadditions.recipes.nuclear.HotCoolantRecipeMap;
 import gregicadditions.recipes.qubit.QubitConsumerRecipeBuilder;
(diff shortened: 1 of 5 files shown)


hjae78 (approved):

hjae78:
Very, very nice!

DStrand1 (changes requested):
There is a lot of duplicate code here, is there not a way in the general recipe map to set an offset for the text depending on the input slots?

TechLord22:
Updated and renamed the two classes. The new classes now both work with any RecipeMap of their equivalent type. The solution for this turned to be pretty simple, which was having a default ProgressBar texture.

DStrand1 (approved):
Seems a bit better, but I still feel like there is a lot of duplicate code. Though maybe this is not something that can really be helped, at least not within scope.

huneau:
> Seems a bit better, but I still feel like there is a lot of duplicate code. Though maybe this is not something that can really be helped, at least not within scope.

yeah I agree

huneau (approved):

hjae78:
addInventorySlotGroup is identical to RecipeMap I think?

Yefancy (approved):

hjae78:
I think we can merge both files and only change `createUITemplate()`. Basically all we are doing is extending the window since JEI adds total energy starting from the bottom of the window as far as I can tell. I don't think there is any type of GUI math needed here

brachy84 (approved):

PR #406, opened 2021-03-15
Fix GTCE Polarizer recipes not getting removed

ALongStringOfNumbers:
Fixes a Typo that prevents LV-EV GTCE Polarizers not having their recipes removed, causing duplicate recipes with the Gregicality recipes.


huneau (approved):
ok thx a lot

Issue #158, opened 2020-09-08
[Suggestion] UHV+ Wireless Power


Shattered2909:
I would love to see completely wireless power transmission at this stage of the game with some kind of large Multiblock Structure to send/receive energy, maybe an "upgraded" Battery Tower? It would have a certain range (maybe add some kind of block in the Multi that has different tiers for range) and Pylons to extend the range in a certain area. At this point pretty much all the machines should be Multiblocks so it would make sense to add Wireless Power Hatches that are connected to the Tower/a Pylon.

Shattered2909:
I would go with some kind of Emitter Block and make this Multiblock take UHV or above.

Zalgo239:
Here's my take on wireless power transfer:

A large tesla coil multiblock, capable of transmitting power to machines (possibly if those are given a special receiver upgrade).

What I think the basic one could look like:
![2020-09-08_20 05 01](https://user-images.githubusercontent.com/67443118/92514225-d6c7b700-f211-11ea-88dd-f47b5fd2fac7.png)

And view from the top:
![2020-09-08_20 25 56](https://user-images.githubusercontent.com/67443118/92514259-e3e4a600-f211-11ea-996d-0306a313b7a7.png)

The base could be made of pretty much any block (an insulator would be preferred), the base supports the primary coil (large circle of coil blocks) and the secondary coil (tall tower wrapped in coil blocks)
The tower itself is made of polyethylene blocks with coils wrapped around and as seen in the top view it is empty
The torus at the top should be made of a conductive and possibly heat resistant block

The coil should transmit power to nearby machines with an efficiency that goes down according to the inverse square law.
To upgrade the transmitted power the player should swap the coils with higher tier ones to transmit an higher total amperage (the sum of all the amps consumed by all machines drawing power from the coil). Swapping the coils could also allow the player to transmit a higher voltage tier
To upgrade the range the player should make the tower higher, realistically the max range should scale with the square root of the height, this would also make extremely tall coils unpractical.

The coil could also have a secondary operational mode, in this mode it is unable to provide power to machines, but it can shock mobs and players(it would need a whitelist/blacklist) from a longer range than the normal tesla coil block. Using higher tier coils deals more damage, and making the tower higher increased the max range like in the other operational mode

Issue #352, opened 2021-02-04
wireless power ?


huneau:
I see 2 topics with the wireless:
- 1 long range wireless 1:1
- 2 short range wireless 1:X

and my idea is more the long range but with a bit of flexibility with tiered wireless that provide more slot  (1 slot = 1 connections)
for example =>
IV tier => 2 slots => MAX 2 amps => 25% loss => 32 block raduis (2 chunks)
LuV tier => 4 slots => MAX 4 amps => 20% loss => 64 block raduis (4 chunks)
ZPM tier => 8 slots => MAX 8 amps => 15% loss => 128 block raduis (8 chunks)
UV tier => 16 slots => MAX 16 amps => 10% loss => 256 block raduis (16 chunks)

and then UHV+ => 1 Slot => MAX 64 amps => 0% loss => unlimited range / dim => consume 1Qubit/t

shift right click with data orb on a tile entity to lock the coordinate and place it in the wireless transmitter

make a cover like a wireless energy accepter to let the energy flow to the machine

brachy84:
Hi, I'm still waiting for your changes and I have a few quations.
 * This is a single block, right?
 * whats the point in having 256 blocks of range but only 16 slots?
 * What do you think about a wireless inventory charger?
    - You would craft items with a wireless receiver upgrade (which will add a "chargable" tag) so the charger knows which items to charge

huneau:
yes, you right, It should be single block. 
I imagine wireless power like draconic evolution wireless. 
The energy in/out (actually batteries make the job) maybe with a cover.
The relay to transfer energy from A to B (here with multiple slots with higher tier) (its the purpose of this topic)
and then the wireless power to power nearby machines (actually not designed) but we can imagine a thing with cover as well

brachy84:
Oh, nice I'm huuuge fan of the draconic evolution energy crystals. I think i can do that.
How it will work:

- Wireless Energy Network I/O block (maybe just battery buffers)
- Energy Network handler block
    - will have slots for data sticks
    - can only send/drain energy to/from Energy Network I/O blocks
- Wireless Network handler is the same exept it can send energy to any block with wireless receiver cover
- maybe a wireless inventory charger to charge items (like armor) in player inventories

brachy84:
Hi again.
May I ask what's wrong with the tesla tower? I thought that it is to op, but your idea of the wireless thing is pretty much the same exept it is not a cool multiblock.

PR #410, opened 2021-03-17
Tesla tower

brachy84:
This will add the Tesla tower multiblock to Gregicality.
This marked as Not ready because it might need some balance and changes in behavior.
It is based on #352 and #158. This multiblock is intended to be gated to at least UHV.

 ### How it works:
* The multiblock looks pretty much the same as the pictures in #158 from Zalgo
* The middle coils can have height between 3 and 128
* The coil height determines the range
  currently the formula is `pow(coilHeight, 0.6) * 4`
  (7.7 blocks at height 3 and 73.5 blocks at 128)
* The coils can only be the new coils from gregicality
* The coil tier determines the Amps
* The tower will try to power machines every 10 ticks
* The tower will scan it's range for `ICoverable's` with a wireless receiver cover
* The cover itself does nothing besides telling the tower to give power
* I don't know how much lag it would cause when the tower would scan the area every second so I made a button in the gui to manually scan
* The torus can be made out of different materials (maybe the materials could have effects on effeciency and range)
* You can also power machines that are out of range or even in another dimension. To to that:
  * Take a data stick use it on a machine and put it in a input bus of the tower
  * The position and dimension will be shown in the tooltip
  * It will require qubits additionally (1 qubit/t per machine out of range)
 
The amount of energy send per pulse needs to be adjusted.
Tell me what you think and how it could be improved.


huneau:
Hum, its a huge work but I have so many things to say.
I will close #158 because its not what I want, I will enhance #352 with my spec.

src/main/java/gregicadditions/machines/multi/MetaTileEntityTeslaTower.java
```diff
@@ -0,0 +1,594 @@
...
+                .where('C', coilPredicate())
+                .where('G', torusPredicate())
+                .where('S', selfPredicate())
+                .where('#', (tile) -> true)
+                //.where('T', itemInputPredicate())
+                //.setAmountLimit('T', 1, 1)
+                .build();
+    }
+
+    public Predicate<BlockWorldState> torusPredicate() {
```
hjae78, line 309: What effect do the different materials have here?

src/main/java/gregicadditions/machines/multi/MetaTileEntityTeslaTower.java
```diff
@@ -0,0 +1,594 @@
...
+        Block[] blocks = new Block[ALLOWED_TORUS_MATERIALS.length];
+        for(int i = 0; i < ALLOWED_TORUS_MATERIALS.length; i++) {
+            blocks[i] = MetaBlocks.COMPRESSED.get(ALLOWED_TORUS_MATERIALS[i]);
+        }
+        return blockPredicate(blocks);
+    }
+
+    public Predicate<BlockWorldState> itemInputPredicate() {
```
hjae78, line 317: This can be removed

src/main/java/gregicadditions/machines/multi/MetaTileEntityTeslaTower.java
```diff
@@ -0,0 +1,594 @@
...
+            if (tile instanceof MetaTileEntityHolder) {
+                if (((MetaTileEntityHolder) tile).getMetaTileEntity() != null) {
+                    if (hasCover(((MetaTileEntityHolder) tile).getMetaTileEntity())) {
+                        blocks.add(new BlockPosDim(pos, getWorld().provider.getDimension()));
+                    }
+                }
+            }
+        }
+        return blocks;
+    }
```
hjae78, line 552: I'm not sure if I understand this, do we feed energy to all block positions given by the data sticks and then scan for machines that have the cover and feed energy to them as well?

src/main/java/gregicadditions/machines/multi/MetaTileEntityTeslaTower.java
```diff
@@ -0,0 +1,594 @@
...
+
+    protected void handleButtonClick(Widget.ClickData clickData) {
+        receivers = scanRange();
+    }
+
+    @Override
+    protected ModularUI.Builder createUITemplate(EntityPlayer entityPlayer) {
+        return super.createUITemplate(entityPlayer);
```
hjae78, line 560: Is this override needed?

src/main/java/gregicadditions/machines/multi/MetaTileEntityTeslaTower.java
```diff
@@ -0,0 +1,594 @@
...
+    };
+
+    private static final SolidMaterial[] ALLOWED_TORUS_MATERIALS = {
+            Materials.BlackBronze,
+            Materials.Electrum,
+            GAMaterials.Pikyonium
+    };
+
+
+    protected final double rangeFactor = 4.0D;
```
hjae78, line 69: This could be moved to config

src/main/java/gregicadditions/machines/multi/MetaTileEntityTeslaTower.java
```diff
@@ -0,0 +1,594 @@
...
+    public long feedEnergy(BlockPosDim pos, long amount) {
+        boolean consumeQubit = false;
+        World world = getWorld();
+        if(!isInRange(pos)) {
+            consumeQubit = true;
+        }
+        if(!isInDim(pos.getDim())) {
+            consumeQubit = true;
+            world = DimensionManager.getWorld(pos.getDim());
+        }
```
hjae78, line 430: Very nice idea!

src/main/java/gregicadditions/machines/multi/MetaTileEntityTeslaTower.java
```diff
@@ -0,0 +1,594 @@
...
+            } else {
+                removeLater.add(pos);
+                failedPositions.remove(sPos);
+                return 0;
+            }
+        }
+
+        if (container != null && container.inputsEnergy(EnumFacing.NORTH)) {
+            if(consumeQubit) {
+                int qubitCost = 10;
```
hjae78, line 455: Another config candidate

hjae78:
Very nice work Brachy! Nice to have you back!

src/main/java/gregicadditions/machines/multi/MetaTileEntityTeslaTower.java
brachy84: It scans all blocks in range for entities with the cover and puts them in the list.

The data stick is only for when your machine is out of range

src/main/java/gregicadditions/machines/multi/MetaTileEntityTeslaTower.java
```diff
@@ -0,0 +1,594 @@
...
+    protected ModularUI createUI(EntityPlayer entityPlayer) {
+        ModularUI.Builder builder = this.createUITemplate(entityPlayer);
+        builder.widget(new ClickButtonWidget(125, 103, 40, 18, I18n.format("gtadditions.multiblock.tesla_tower.scan"), this::handleButtonClick));
+        return builder.build(this.getHolder(), entityPlayer);
+    }
+
+    @Override
+    protected void handleDisplayClick(String componentData, Widget.ClickData clickData) {
+        super.handleDisplayClick(componentData, clickData);
+    }
```
DStrand1, line 573: Same as above, this override is unneeded in current impl

DStrand1 (changes requested):

DStrand1:
I am changing this PR to draft instead of stating "Not Ready" in the title

src/main/java/gregicadditions/machines/multi/MetaTileEntityTeslaTower.java
brachy84: currently none. But the tower could have a default energy loss of 40%.
The blocks then could have stats like conductivity (which would decrease energy loss) and a range modifier
A maxed out tower should be made out of UXV Superconductor blocks to have 0-1% loss and maximum range.
None of these are implemented yet. Just ideas. I could also add a crafttweaker method to add torus blocks

brachy84:
I thought about this a little more and I think I will close it. Because the tower is really op. Yeah, it's pretty cool, but not needing any cables at UHV and above is just game breaking. So I will probably close this and maybe make something that fits more what decal wants and is more balanced. Unless we find a good way to balance this.

PR #413, opened 2021-03-18
Electric Implosion Compressor

DStrand1:
Adapted from Bartworks, the **Electric Implosion Compressor** is a multiblock that serves the same purpose as the Implosion Compressor, but is significantly faster and does not use any form of explosives in its recipes.

## Differences from BW
In BW, the Electric Implosion Compressor was a 3x9x3 multiblock, taking a whopping 34 Blocks of Neutronium to craft and form. In my implementation, the Neutronium is swapped to Americium, since it is still relatively hard to get, but not nearly as difficult. Additionally, I made the machine only 3x5x3, and trimmed out some unnecessary parts in the multiblock structure.

This multi was UEV-tier in BW. I changed this to UHV for us, to make it a little easier to use, since it is not nearly as important in Gregicality as in GT5. I also cheapened the recipe to craft the Controller by quite a bit, though it is still fairly expensive.

## Recipe Generation
Similar to how I implemented the recipes for the Disassembler, I based the Electric Implosion Compressor's recipes off of the Implosion Compressor's RecipeMap. I ensured that the explosive was ignored from the new recipe, even checking for the property for explosives to ensure that any explosive will be removed. Additionally, I made sure that recipes would not be duplicated, as we have recipes with TNT and Dynamite for the Implosion Compressor.

Note that this recipe generation does NOT take not-consumed items into account, but since this code runs before CraftTweaker recipe addition, it will be fine as long as we do not add any recipes using circuits or other not-consumed items to the Implosion Compressor's RecipeMap. If this needs to be changed, I can, but I did not feel it was necessary for now.

## Gallery
The multiblock formed, taking a mix of Steel and Americium Casings:
inWorld

The recipe, using 10 ingots worth of Osmium, Americium, and Soldering Alloy fluids, with Osmium Rings and Long Rods, Americium Blocks, UV Pistons and Superconductor Wire, and a normal Implosion Compressor Controller:
recipe

The JEI preview:
jeiPreview

An example of a Recipe:
exampleRecipe

### DIFF
--- a/src/main/java/gregicadditions/machines/multi/MetaTileEntityElectricImplosion.java
+++ b/src/main/java/gregicadditions/machines/multi/MetaTileEntityElectricImplosion.java
@@ -10,7 +10,6 @@ import gregtech.api.metatileentity.multiblock.MultiblockAbility;
 import gregtech.api.multiblock.BlockPattern;
 import gregtech.api.multiblock.FactoryBlockPattern;
 import gregtech.api.render.ICubeRenderer;
-import gregtech.common.blocks.MetaBlocks;
 import net.minecraft.block.state.IBlockState;
 import net.minecraft.util.ResourceLocation;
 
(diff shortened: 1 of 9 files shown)


src/main/java/gregicadditions/machines/multi/MetaTileEntityElectricImplosion.java
```diff
@@ -0,0 +1,58 @@
...
+    @Override
+    public MetaTileEntity createMetaTileEntity(MetaTileEntityHolder metaTileEntityHolder) {
+        return new MetaTileEntityElectricImplosion(metaTileEntityId);
+    }
+
+    @Override
+    protected BlockPattern createStructurePattern() {
+        return FactoryBlockPattern.start()
+                .aisle("XXX", "AAA", "AAA", "AAA", "XXX")
+                .aisle("XXX", "A#A", "A#A", "A#A", "XXX")
```
hjae78, line 37: I think piston blocks would be good to add replacing the top and bottom empty slot

hjae78:
Nice work 👍

huneau:
not sure to see the need for that multiblock. the implosion compression has mostly no use ....

DStrand1:
I like the Implosion Compressor multiblock a lot, so I was hoping that this would encourage more recipes to make their way to it

src/main/java/gregicadditions/recipes/GARecipeMaps.java
```diff
@@ -164,5 +167,8 @@
                 .setSlotOverlay(true, false, GuiTextures.CIRCUIT_OVERLAY)
                 .setProgressBar(GAGuiTextures.PROGRESS_BAR_DISASSEMBLER, ProgressWidget.MoveType.HORIZONTAL);
 
+        ELECTRIC_IMPLOSION_RECIPES = new RecipeMap<>("electric_implosion", 1, 2, 1,
+                2, 0, 0, 0, 0, new SimpleRecipeBuilder()
+                .duration(1).EUt(1500000)); // UHV-tier, 1tick processing time
```
TechLord22, line 172: Making the EUt a config option might be helpful for packdevs, etc.

TechLord22 (approved):

hjae78:
I really like the idea of this but maybe we should also think of some new material which is required in large quantities to incentivize it's use. Perhaps a new gem? Or even some sort of new advanced plating like Iridium Alloy Plating or whatever it's called

hjae78 (approved):

pablocnvs (approved):
I do agree on Decal's point that it isn't used that much, having the Stellar Forge really limits the additions to the implosion compressor. For the same reason by UHV you would like to prioritize the glycerly nitrate for the stellar forge charges, so i feel it is a solid addition.

Zalgo239 (approved):
This would be nice to have as a powerful but explosive-free compressor, there are probably some things that would make more sense to be made in this, I'll have to take a look at the chains we have to see what makes sense to move over.

Yefancy (approved):

PR #414, opened 2021-03-18
Fix Insulation Wire Assembly Recipes

DStrand1:
Fixes an issue where wire recipes using Insulation Wire Assembly had bad recipe amounts.

Since Insulation Wire Assembly is expensive, the recipes were trying to follow the pattern:
`wireInsulationAmount = (cableAmount == 1) ? 1 : cableAmount / 2`
to save on resources, but was accidentally not doing this, and changing the cable amount too. It now follows this ternary pattern.

Old:
old

New (1x):
1x

New (2x):
2xpt1

New (4x):
4x

This is not an inclusive list of the recipes, but they are all fine now, and these screenshots show the pattern.


pablocnvs (commented):
One thing i was trying to do is reduce the amount of Zylon and Fullerene Polymer Matrix, since the  no longer serve an insulating function as much. If you think 1 foil for 4x cables is too low you can keep it, though.

hjae78 (approved):

Yefancy (approved):

PR #423, opened 2021-03-23
Equiptment config & Nightvision goggles

brachy84:
This will add configs for armors, jetpacks and batpacks. It also adds Nigthvision goggles. By default they are set to MV. They also give no protection. 
The nano suit now extends nightvision goggles to avoid duplicate code. The crafting recipe for the nano helmet changed also. It needs less components now, but the nigth vision goggles
(shitty texture by me)
grafik


hjae78 (approved):
Very nice! Been on my todo list for ages. If you would like you could add proper names and perhaps even comments (although that is quite unnecessary) for the config values.

DStrand1 (approved):
Nice! Love the configurability of it all

TechLord22 (approved):
Looks great!

PR #426, opened 2021-03-24
fix PA not running due to searching for recipemap with localised string

PrototypeTrousers:
During testing some improvements to recipe logic, i stumbled this issue.
for example, a PA with an macerator would look for "Pulveriser" as string and not the proper "macerator" string that we ca retrieve the recipemap.
Also, after the recipe was built it would add the machine as an input of the recipe, due to leftover of the old logic when the machine stack worked from inside the bus.

removed adding machine stack to the recipe when setting the machine stack to not consumable.
removed reflection.
removed comparison by string references


hjae78:
How come you closed this?

PrototypeTrousers:
i messed up and the comments didnt match the code changes.

PR #427, opened 2021-03-24
fix PA not running due to add the machinestack as recipe input

PrototypeTrousers:
supersedes #426

After the recipe was built it the PA adds the machine in the dedicate slot as an input of the recipe, due to leftover of the old logic when the machine stack worked from inside the bus.

removed adding machine stack to the recipe when setting the machine stack to not consumable.
removed reflection.
removed comparison by string references


DStrand1 (commented):
Looks good! Still don't know how that MachineStack as a notConsumed input was still in there...

DStrand1 (approved):

PR #429, opened 2021-03-27
Battery Overhaul

pablocnvs:
This PR overhauls the production of batteries, replacing the old batteries with newer ones based on current technologies and make their production be more in line with the current style of the game. The pr has been coded to ensure nobody loses the batteries they have made, so the new batteries relate to the old ones. Expect however a shift in textures, as the previous material tiers (Vanadium, Naquadria and Neutronium) have been replaced by tiers based on cell chemistry (Lithium Ion, Lithium Sulfide and Fluoride).
As a bonus, the pr gives more uses to items and fluids that have just one use, encouraging the investment in those chains.

PS: The UMV and UXV wires are a bit of a placeholder, as there is no more materials for those tiers


brachy84:
If I'm not mistaken then this will delete all existing batterys from existing saves, which is obviously not good

pablocnvs:
I'm going to retest that, but i think first time everything went ok

pablocnvs:
Proof that no batteries are lost:





Textures have changed, but tier remains the same

hjae78 (approved):

Zalgo239 (approved):

Yefancy (approved):

DStrand1 (approved):
Looks good! At this point though, I feel we should maybe rename or even remove the config option for "GT5u ZPM/UV batteries" since they are not from GT5U anymore.

PR #433, opened 2021-03-31
Move Recipe Reversal Recipes to After CT

DStrand1:
This PR is simple, and moves both the Electric Implosion Compressor and Disassembler recipe generation calls to be after CraftTweaker recipes have been generated. This way, these two machines will properly respect the updated recipes instead of needing to be manually changed for all CraftTweaker changes implemented.

This is extremely straightforward for the Electric Implosion Compressor, but for the Disassembler, it may be unwanted, so there is still always the config option for it.

Secondly, I fixed an issue with the Electric Implosion recipe generator where it did not respect the stack size of the input ingredient.

### DIFF
--- a/src/main/java/gregicadditions/CommonProxy.java
+++ b/src/main/java/gregicadditions/CommonProxy.java
@@ -58,6 +58,7 @@ public class CommonProxy {
     }
 
     public void onLoad() throws IOException {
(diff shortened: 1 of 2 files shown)


DStrand1:
I have added this, so this PR is ready for review and merge

PR #434, opened 2021-04-01
Remove scheduleRenderUpdate() method calls

TechLord22:
Potentially prevents a race condition on world load causing an NPE. This should fix it.


src/main/java/gregicadditions/machines/multi/simple/TileEntityLargeBenderAndForming.java
```diff
@@ -4,6 +4,7 @@
 import gregicadditions.GAConfig;
 import gregicadditions.capabilities.GregicAdditionsCapabilities;
 import gregicadditions.capabilities.IMultiRecipe;
+import gregicadditions.client.ClientHandler;
```
DStrand1, line 7: 🤔

DStrand1 (approved):

PR #437, opened 2021-04-02
Mainframe resprite

trerri:
See title.

circuit supracausal mainframe
mainframe wetware
circuit bioware mainframe
circuit cosmic mainframe
circuit mainframe crystal
circuit mainframe micro
circuit mainframe nano
circuit mainframe optical
circuit mainframe quantum
circuit mainframe refined


DStrand1 (commented):
Personally I am not a fan of changing these textures in any way. They take from past versions of GT, and as GTCE is primarily a rewrite of GT5 for 1.12.2, I don't see why these should be changed in the mod.

trerri:
Counter argument: the original ones look bad, and these ones look better (subjective) as they're more consistent with the MC style. Hence, they should be changed.

hjae78:
The mainframe sprite is very old and I think that this works better as a resource pack

hjae78:
I'm closing this for now

PR #388, opened 2021-03-01
Update Chemistry for consistancy with GTCE

TechLord22:
## Introduction
This PR addresses issues in every single chain we have in the mod. We have been thorough and made sure to double or triple check every single recipe we have added to the game. Most of these adjustments were to adjust mole amounts, which I will get into later, though there was a fairly substantial amount of recipes reworked to maintain proper chemistry and avoid positive loop exploits/lossy recipes.

## Conventions
As per conventions established and followed rigorously by GTCE, our mole counts follow two rules:
- Fluids are always 1B per mole
- Dusts are always* N dusts per mole, where N is the number of total elements in the compound

See https://github.com/GregTechCE/GregTech/pull/1414 for the original unification of the conventions we are maintaining in this PR.

*: Some dusts have *very* large formulas, and some are almost completely unknown, so in these rare cases, we left them at 1 dust per mole. This is a bad thing to do, but sometimes there is just no other way. Any time this was done, it was handled very carefully to ensure no side effects.

## Recipe Map Changes
As a result of this PR, we have changed the RecipeMaps for the LCR and Chemical Plant slightly, changing input/output amounts.
- **LCR:** Added 1 additional item output
- **Chemical Plant:** Added 2 item inputs and 2 item outputs
- **Large Engraver:** Added 1 fluid input and 1 fluid output

This allowed for some of our more complex balanced reactions to fit in a machine, and also helps differentiate the Chemical Plant from the LCR. Additionally, it will help for the future when condensed processes are eventually added to the Chemical Plant.

## Material Changes
Since fluid tooltips now exist, we have implemented a way for tooltips to be provided for fluids. We have this provided as either an ImmutableList like before, or simply as a String. There is also now a way to give a tooltip to a SimpleDust as a String instead of an ImmutableList. This should be the preferred way, since it is more efficient on load time since we were not using the ImmutableList for anything other than tooltip generation anyway.

There was a fair amount of materials changed, mostly due to misspelled variable names, translation keys, and translated names. These have all been updated, but there may be some warnings on updating. As a result, the release including this PR should definitely be stated in the release notes very clearly.
\
\
\
**Credits**
- @TechLord22 For doing the first pass of all recipes and the majority of the work
- @DStrand1 For doing the second pass of all recipes, implementing the new tooltip systems, and writing this opening comment.
- The rest of the Gregicality devs, for helping with balancing formulas and determining the formulas of some compounds (and adding the original recipes of course)

**Issues Closed**
Closes #371
Closes #392
[...]

PR #439, opened 2021-04-05
[Tool] Freedom Wrench

Yefancy:
### Freedom Wrench
An interesting tool:

- Allows the player to rotate the multiblock controller (such as sideways structure).
- Projection display the structure (reference from JEI).
- Debug mode helps check the position of the error block.

video: [youtube](https://www.youtube.com/watch?v=la4ATNYrzEo&list=PLBAo7z4aqr3-GwXkilU4BVXHDErygdvMt&index=5&ab_channel=KilaBashLiang)

### Image:
image
image
image


hjae78:
Wow, just amazing work as always

huneau:
yes awesome job, the last thing it needs its a building mod like environmental tech

src/main/java/gregicadditions/coremod/GAClassTransformer.java
```diff
@@ -41,6 +42,12 @@
             case "gregtech.api.util.GTUtility":
                 tform = GTUtilityTransformer.INSTANCE;
                 break;
+            case "gregtech.api.multiblock.BlockPattern":
+                tform = BlockPatternTransformer.INSTANCE;
+                break;
+            case "gregtech.api.metatileentity.multiblock.RecipeMapMultiblockController":
+                tform = RecipeMapMultiblockControllerTransformer.INSTANCE;
```
hjae78, line 49: You reference `RecipeMapMultiblockController` a lot, does this not work for multiblocks that aren't of this instance? For example Large Miners, Drilling Rig

Yefancy: projection work for all `MultiblockControllerBase`. but rotation just work for `RecipeMapMultiblockController` at the present. 
After rotation, BlockPattern detection and rotation of the overlay are working well, but i have noticed that some machines may have strict requirements for `frontFacing`, which must not be facing up or down. (e.g. [Cosmic Ray Detector](https://github.com/Gregicality/gregicality/blob/5bf0eb12f3a6bf8dbb7e4627189219434e52fb00/src/main/java/gregicadditions/machines/multi/MetaTileEntityCosmicRayDetector.java#L85-L90))

DStrand1:
At this point, I feel like at least the in-world preview should be moved up to GTCE, though it is already done by Multiblock Tweaker. Have you tested what happens with both mods installed?

Yefancy:
> At this point, I feel like at least the in-world preview should be moved up to GTCE, though it is already done by Multiblock Tweaker. Have you tested what happens with both mods installed?

- I implemented a feature similar to Multiblock Tweaker because In-World Preview is required to support freedom rotation (I got Eutro's support for this). I don't want to implement projection for GTCE, and rewrite a freedom rotation supporting one for Gregicality. It is better not to separate the two works.
- They work well together, or at least don't cause crash,

DStrand1:
I have been wanting to implement something like MBT’s in-world preview into GTCE, so I will see if we can get this full PR into GTCE. I can assist with porting it to there if it gets approved.

Yefancy:
> I have been wanting to implement something like MBT’s in-world preview into GTCE, so I will see if we can get this full PR into GTCE. I can assist with porting it to there if it gets approved.

sure, it would be nice if it could be integrated into GTCE. I don't want to do hacking gtce all the time.

src/main/java/gregicadditions/GAConfig.java
```diff
@@ -1393,7 +1393,7 @@
             @Config.RequiresMcRestart
             @Config.Name("Prospector scan cost")
             @Config.RangeInt(min = 0)
-            public int[] scanCosts = {32, 138, 8192, 32768};
+            public int[] scanCosts = {8, 32, 512, 2048};
```
DStrand1, line 1396: Seems there are a few commits jumbled into here as a result of you needing the GTCE 1.14.0 crash fix. Since that is included in master, dropping those commits and rebasing your branch should be safe and would help to not touch files that are not part of this PR.

src/main/java/gregicadditions/coremod/transform/GTUtilityTransformer.java
```diff
@@ -1,11 +1,12 @@
...
 
-import gregicadditions.coremod.GAClassTransformer;
+import gregicadditions.coremod.GAClassTransformer.ClassMapper;
+import gregicadditions.coremod.GAClassTransformer.GAMethodVisitor;
 import org.objectweb.asm.ClassVisitor;
 import org.objectweb.asm.MethodVisitor;
 import org.objectweb.asm.Opcodes;
 
-public class GTUtilityTransformer extends GAClassTransformer.ClassMapper {
+public class GTUtilityTransformer extends ClassMapper {
```
DStrand1, line 9: The change to this class is unnecessary since I am removing it as part of #388. If possible, can you revert the change to just this class? Otherwise either this PR or the Chemistry PR will have to rebase due to this file changing in master, depending on which PR is merged first.

DStrand1 (changes requested):
Rendering really isn't my thing, so my review is more focused on some logistics of this PR being merged more than anything else. Hope it isn't too annoying

src/main/java/gregicadditions/coremod/transform/GTUtilityTransformer.java
Yefancy: you are right, i noticed that this PR is a bit of a mess after rebase. i will continue to work here, but this PR will always be WIP until it is closed. after that, I will create a new branch and PR after Chemistry PR merged.

PR #442, opened 2021-04-07
Fix Prospector Hard Crash

DStrand1:
Some of the parameters for the ProspectorStats constructor were out of order, leading to the scanCosts field being set to the chunk radius. This would cause the server instance of the world to get stuck trying to scan a chunk radius in the thousands, leading to a hard crash (and sometimes world corruption) on force close.

Additionally, I decreased the scanCosts even further since with default settings, the Prospector still could not be kept open for more than a few seconds from full charge with default settings.

Personally, I think that this highlights an issue of commits being applied directly to master. I think that we should have all code changes (except for the occasional compilation error fix) come through a PR first so that at least trivial typos can be caught by the rest of the development team.

### DIFF
--- a/src/main/java/gregicadditions/GAConfig.java
+++ b/src/main/java/gregicadditions/GAConfig.java
@@ -1393,7 +1393,7 @@ public class GAConfig {
             @Config.RequiresMcRestart
             @Config.Name("Prospector scan cost")
             @Config.RangeInt(min = 0)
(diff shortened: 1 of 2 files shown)


Yefancy (approved):

TechLord22 (approved):

hjae78:
You're right, I'll pr my commits from now on

PR #446, opened 2021-04-11
Fix Assorted Multiblock Problems

### LINKED ISSUE #111: Tumbaga EBF time
**Tumbaga**, which is required for LV Integral Framework, currently takes **192 seconds** per ingot in the EBF.
image

160 Tumbaga is needed for the **Chemical Plant**, which translates to **8.53 hours** for 1 EBF running on MV power, or just over **2 hours** for 4.

Tumbaga's recipe seems excessively expensive compared to other alloys needed for Integral Frameworks, such as:
Eglin (MV) - 49.25s at MV
Inconel 625 (HV) - 155.2s at MV

TechLord22:
This PR fixes and updates many different multiblock machines.

I'll start with detailing the more significant changes below:

First, all multiblocks displaying information such as EUt do not show it more than once. Additionally, multiblocks that did not display the "Not Formed" text now do. 

Second, multiblocks like the large forge hammer, large washer, volcanus, and cryogenic freezer now display the correct multiblock info in their tooltips. They previously had info sourced from the lang, which did not reflect the config.

Third, the Advanced Distillation Tower was one of the strangest multiblocks we have had. It required the IV-tier integral framework in it's structure, and was built to be 26 blocks tall, even though recipes would not even output beyond layer 12. This has now been changed to be built identically to the regular Distillation Tower, except with different casings and tiered integral frameworks instead. This will make the multiblock much more viable, and worth using.

Fourth, The Industrial Primitive Blast Furnace's structure is now different. The output bus is now in the back center, and the input bus can be on either side of the multiblock, making automation more flexible. Additionally, and most notably, the multiblock does not require a bus on every slice. This makes it much more viable since its prior cost in buses is far higher than it should have been. 

Fifth, many English lang entries have been updated with corrected spelling, grammar, and information to reflect changes made in this PR and in the past.

Minor Changes:
1. The Chemical Plant recipe has been changed to use Cobalt Brass plates, instead of dust. Dust made no sense, and Cobalt Brass is smelted in a regular furnace, so any game balance should not be affected here.
2. The LV integral framework recipe has been changed from using Tumbaga to Magnalium. Tumbaga does indeed take far too long for an LV-tier component, so Magnalium was chosen instead. Although it has fewer steps to make than Tumbaga, it is better suited as an LV-tier material, since it is made in the alloy smelter instead of the ebf. This removes the massive time-gate for the chemical plant. Closes #111.
3. The Hyper Reactors now say the material used to boost them in their tooltips.
4. The Chunk Miner's fluid GUI text is properly distanced from the top of its section.
5. The Large Packager is now IV tier instead of LuV. Players felt it was too expensive in it's current state since it was gated behind the assembly line. As a result, HG-1223 has been moved to HSS-G coils instead of naquadah coils. The recipe has been changed to use an IV hull and components.


DStrand1 (approved):

src/main/java/gregicadditions/machines/multi/advance/hyper/HyperReactorUEV.java
```diff
@@ -89,6 +88,7 @@ protected void addDisplayText(List<ITextComponent> textList) {
     @Override
     public void addInformation(ItemStack stack, @Nullable World player, List<String> tooltip, boolean advanced) {
         super.addInformation(stack, player, tooltip, advanced);
+        tooltip.add("Boost With: " + Helium.getMaterialPlasma().getLocalizedName(Helium.getFluid(1)));
```
hjae78, line 91: This should take into account the config set fluid

hjae78 (changes requested):

huneau:
Im not a huge fan of the changes from IPBF. it seems too powerful actually.
The reason for this row of buses is to avoid using it instead of the blast furnace.

pablocnvs:
I do understand your doubts about the change, but i think the limited recipemap of the IPBF kind of balances out this. Using coils will be inevitable for progression if we give the alloy furnace coil requirements as well. Maybe early game it will be powerful but by HV it should have died down.

TechLord22:
I also think that the fuel consumption might also make it less energy efficient than using a large boiler, etc. When testing with a 64 block long IPBF, it used coal coke blocks every ~1.5s. No one will be able to sustain this, and if you were able to, it would be far more effective to turn into electricity instead. I think that is another important tradeoff that balances out the multiblock further.

TechLord22:
This most recent commit addresses a GTCE bug where the distillation tower max height (and thus outputs) is actually 11 instead of the RecipeMap's 12.

hjae78:
I'm fine with the IPBF changes, considering how much we've extended the late game I see no reason for the early game not to be sped up even if just a little bit

hjae78 (approved):

hjae78:
Would you like changes @huneau or okay to merge?

pablocnvs (approved):

PR #449, opened 2021-04-13
Remove the HWYLA maven

ALongStringOfNumbers:
Removes Tehnut's maven as nothing in the project was depending on it. Required making Ex Nihilo Creatio not transitive so it did not attempt to bring HWYLA with it

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -49,9 +49,6 @@ repositories {
         url = "http://maven.ic2.player.to"
     }
     maven {
-        url "https://maven.tehnut.info/"
-    }
-    maven {
         url "https://minecraft.curseforge.com/api/maven"
     }
     maven {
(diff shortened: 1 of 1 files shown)


TechLord22 (approved):

build.gradle
```diff
@@ -100,7 +97,9 @@ dependencies {
     provided "li.cil.oc:OpenComputers:MC1.12.2-1.7.5.192"
     provided "appeng:appliedenergistics2:rv6-stable-7"
     provided "binnie:binnie-mods-1.12.2:2.5.1.184"
-    provided "exnihilocreatio:exnihilocreatio:1.12.2-0.3.7.31"
+    provided ("exnihilocreatio:exnihilocreatio:1.12.2-0.3.7.31") {
+        transitive = false
```
hjae78, line 101: How come this is needed?

DStrand1: Transitive means that exnihilo will try to pull its own dependencies into our environment when the project is run. This leads to us needing HWYLA’s maven, because it is the only dependency that needs it. With transitive false, HWYLA is no longer pulled, so the maven isnt needed

ALongStringOfNumbers: Ex Nihilio Creatio depends on HWYLA, and so was attempting to transitively pull HWYLA into the development environment, leading to issues when the maven declaration was not present. I believe that it is only mods that are `provided` that attempt to transitively pull in dependencies, although I could be mistaken.

PR #450, opened 2021-04-14
[Tool] Freedom Wrench

Yefancy:
Basically the same as #439 

Additional Contents:
- Central monitor can be rotated at will now
- Supporting rotation for all multi-block now. (some machines aren't perfect, but don't crash)
- Building mode: fast build (no consumption in creative mode)

GIF:
rotation

wrench


hjae78:
Fucking amazing! Building mode finally!

Yefancy:
yeah, Decal is right that this feature is needed. Make testing in creative mode easier

Zalgo239:
This is awesome! Will save lots of time

htmlcsjs:
Let's fucking go

hjae78 (approved):

src/main/java/gregicadditions/item/behaviors/FreedomWrenchBehaviour.java
```diff
@@ -0,0 +1,203 @@
...
+                        facing = side.getOpposite();
+                    }
+                }
+                if (rotateSpin) {
+                    EnumFacing next = BlockPatternChecker.getSpin(mte).rotateY();
+                    if (!world.isRemote) {
+                        BlockPatternChecker.setSpin(mte, next);
+                    } else {
+                        if (facing != EnumFacing.DOWN && facing != EnumFacing.UP) {
+                            player.sendMessage(new TextComponentString("Spin: " + (next == EnumFacing.NORTH ?
```
DStrand1, line 156: I think these should be moved to translation keys, and it can simply be changed from `TextComponentString` to `TextComponentTranslation`. There are a few below this as well. With this change, I think this needs to be within a `world.isRemote` check, though it may need one even without the change

DStrand1 (changes requested):
A lot of this code is rendering and packet handling, most of which is hard to review just from code and I haven't found errors in testing so I will trust it. However, has this been tested on a dedicated server instance? I have some concerns about certain sections, especially in `FreedomWrenchBehavior` not testing for client or server instance everywhere it should be. Other than that, I think this is ready to go

src/main/java/gregicadditions/item/behaviors/FreedomWrenchBehaviour.java
Yefancy: thanks for your review. i have test it on a dedicated server instance before, and it works well. `FreedomWrenchBehavior` do side checking [here](https://github.com/Gregicality/gregicality/pull/450/files/8c49694f695858944423bae184f9777c2294adab#diff-2c8c8977a3f6d21fd5422accd4a1a931f9c36ae250724007165d6b1ddac5b18aR43). 

ill add translation keys later

PR #451, opened 2021-04-15
Distinct Input Buses

DStrand1:
This PR does 1 thing: it adds a button to the UI of a Large Multiblock (anything that extends `LargeSimpleRecipeMapMultiblockController`) that toggles between "distinct" and "combined" item buses.
image

What this means is, if the multiblock is in "combined" mode, it will behave just as it is now, treating all of the Input Buses as one large inventory for recipe matching and item consumption. If it is in "distinct" mode, each Input Bus is completely separate from others. 

For example, if I have a Plate Extruder Shape in Input Bus 2 on my Large Extruder, and Steel Ingots in both Input Buses 2 and 3, then the machine will consume all of the Steel in Input Bus 2, creating Plates, and will not touch any of the Steel in Input Bus 3.

Another example. If I have a Plate Shape in Input Bus 2, and a Ring Shape in Input Bus 3, then any Steel put into Input Bus 2 will create Plates, and any Steel put into Input Bus 3 will create Rings.

The hover tooltip for the UI button:
image

This option is available for all Large Multiblocks, and is set to false by default to retain current behavior on update for players.


htmlcsjs:
This would be very helpful in automation of things that use molds, say with ae2

hjae78 (approved):

huneau:
> This would be very helpful in automation of things that use molds, say with ae2

actually, this works perfectly. there is a default behavior which scans distinct first then combine
but the PR avoid the Combine behavior that can lead to an unwanted recipe.

src/main/java/gregicadditions/machines/multi/simple/LargeSimpleRecipeMapMultiblockController.java
```diff
@@ -216,16 +231,48 @@ public boolean checkRecipe(Recipe recipe, boolean consumeIfSuccess) {
     protected void addDisplayText(List<ITextComponent> textList) {
         super.addDisplayText(textList);
         textList.add(new TextComponentTranslation("gregtech.multiblock.universal.framework", this.maxVoltage));
+
+        ITextComponent buttonText = new TextComponentTranslation("gtadditions.multiblock.universal.distinct");
+        buttonText.appendText(" ");
+        ITextComponent button = withButton((isDistinct ?
+                new TextComponentTranslation("gtadditions.multiblock.universal.distinct.yes") :
```
huneau, line 238: I dont understand why I used `universal` instead of `common` -_-

src/main/java/gregicadditions/machines/multi/simple/LargeSimpleRecipeMapMultiblockController.java
```diff
@@ -285,24 +344,119 @@ protected void trySearchNewRecipe() {
...
-            List<IItemHandlerModifiable> itemInputs = ((RecipeMapMultiblockController) this.getMetaTileEntity()).getAbilities(MultiblockAbility.IMPORT_ITEMS);
+        // Distinct buses code =========================================================================================
+
+        private void trySearchNewRecipeDistinct() {
+            long maxVoltage = getMaxVoltage();
+            Recipe currentRecipe = null;
+            List<IItemHandlerModifiable> importInventory = getInputBuses();
+            IMultipleTankHandler importFluids = getInputTank();
 
-            Tuple<Recipe, IItemHandlerModifiable> recipePerInput = itemInputs.stream()
```
huneau, line 292: you rewrite lots of code, but the only thing to change is that single line

huneau:
otherwise nice feature

src/main/java/gregicadditions/machines/multi/simple/LargeSimpleRecipeMapMultiblockController.java
DStrand1: Its possible that it would be "functional" with a single change there, but there would be many places where the code is largely inefficient. I rewrote `trySearchNewRecipe` and `checkRecipeInputsDirty` because when the buses are treated distinctly, the caching for these methods will fail every single time. But with this change, it properly caches and will prioritize continuing with the same bus rather than swapping to other buses randomly.

hjae78:
@huneau for me this seems okay and from Dan's testing seem to improve the behavior, is it okay to merge?

Yefancy (approved):

PR #455, opened 2021-04-16
Buff assembler component recipes

DStrand1:
This PR simply makes all assembler component recipes for [motors, pistons, robot arms, conveyors, sensors, emitters, and pumps] to be **60V, 5s** for all recipes. This makes them more desirable for saving slight amounts of materials vs before, where it consumed a large amount of power, and took 30 seconds.


hjae78:
I'm not sure I agree with lowering the EU/t that much... I will wait to hear from others before I merge

TechLord22:
My thoughts:
For the assembler, make the EUt equal to the component's tier. So LV motors use 32EUt, HV uses 512, etc. I would lower the duration here to 5 seconds. This keeps these recipes reasonable for the tier you unlock them at. 

For the assembly line, make the EUt equal to the component's tier just as above. Having the duration at 10s will make the assembly line 2x slower than the assembler, but this balances it out because of the decreased material usage in the assembly line.

With these, the assembler recipes are fast enough to be worth doing, but still makes the assembly line recipes a worthwhile upgrade.

DStrand1:
The idea is to make them reasonably overclockable. As it stands, the savings from these recipes are still such that most players probably won’t do these recipes, and opt for the standard crafting table instead due to having to craft so many rounds. But, at least they are more viable now

hjae78:
During my playthrough I never considered the energy cost of these recipes and always used the assembler recipe. Maybe we should even remove the hand crafting recipe for higher tier components (EV/IV)? 
I agree that the voltage should be tiered but I see your point. We could do tiered voltage divided by 2?

DStrand1:
I will redo this PR once #478 is merged.

PR #458, opened 2021-04-22
Temporary fix for FG issues

DStrand1:
Due to changes with Forge's maven hosting locations, and the Forge Team's slowness/refusal to update FG2.3, this PR provides a temporary fix so that we can continue development for now.

The build.gradle is pointed to a forked copy of FG 2.3, fixing a bug that was attempted to be fixed on the official repo, but was refused. If we are satisfied with this solution, we can move the FG 2.3 fork into the Gregicality team, though I think we should wait for a while before doing that.

Additionally I moved AE2 to a provided jar, as for whatever reason it was having issues being fetched after making these changes.


Yefancy (approved):

Yefancy:
nice fixes. but i do also think we should wait a little while

DStrand1:
It is quite clear to me that Forge has no intention of committing fixes to ForgeGradle 2.3, as their "support for old versions" is focused at trying to make ForgeGradle 3 more usable, but that is a large change that I don't think is worth making. This at least lets us continue developing for now so we do not have to wait and see if Forge decides to be nice to us

Yefancy:
I have not noticed forge's intention. if so like you said, i think its really not necessary to wait for forge's feedback

build.gradle
```diff
@@ -87,7 +90,7 @@ dependencies {
     provided "com.github.mcjty:xnet:1.12-1.8.2"
```
Yefancy, line 90: I have a suggestion, I think we should adjust all non-essential dependencies to provoided. Only compile gtce, jei, top, codechicken, CT.

Yefancy, line 90: I have a suggestion, I think we should adjust all non-essential dependencies to provoided. Only compile gtce, jei, top, codechicken, CT.

DStrand1: I agree, this will save a lot of launch time and let us still develop using their code

PR #462, opened 2021-04-28
Buff Nuclear, UU rework

hjae78:
Some buffs to the nuclear process to ease the pain until the proper rework. UU rework according to Zalgos chart. I plan on doing a multiblock for both Replicator and Mass Fab


src/main/java/gregicadditions/GAMaterials.java
```diff
@@ -931,6 +931,9 @@
     public static final SimpleFluidMaterial Cyclopentadiene = new SimpleFluidMaterial("cyclopentadiene", Cyclooctadiene.rgb, "C5H6");
     public static final SimpleFluidMaterial ChloroauricAcid = new SimpleFluidMaterial("chloroauric_acid", 0xDFD11F, "HAuCl?");
     public static final SimpleFluidMaterial Helium4 = new SimpleFluidMaterial("helium4", Helium.materialRGB-10, "He_4");
+    public static final SimpleFluidMaterial FermionicUUMatter = new SimpleFluidMaterial("fermionic_uu_matter", UUMatter.materialRGB / 3, TextFormatting.OBFUSCATED + "???");
+    public static final SimpleFluidMaterial BosionicUUMatter = new SimpleFluidMaterial("bosionic_uu_matter", UUMatter.materialRGB - FermionicUUMatter.rgb, TextFormatting.OBFUSCATED + "???");
```
Zalgo239, line 935: Bosonic is misspelled as Bosionic

Zalgo239 (changes requested):
Bosonic is misspelled as Bosionic, but other than that everything seems to be in order.

src/main/java/gregicadditions/machines/GATileEntities.java
```diff
@@ -773,38 +773,24 @@ public static void init() {
 
         }
 
-        MASS_FAB[0] = create(2175, new SimpleMachineMetaTileEntity(location("mass_fab.lv"), GARecipeMaps.MASS_FAB_RECIPES, ClientHandler.MASS_FAB_OVERLAY, 1));
-        MASS_FAB[1] = create(2176, new SimpleMachineMetaTileEntity(location("mass_fab.mv"), GARecipeMaps.MASS_FAB_RECIPES, ClientHandler.MASS_FAB_OVERLAY, 2));
-        MASS_FAB[2] = create(2177, new SimpleMachineMetaTileEntity(location("mass_fab.hv"), GARecipeMaps.MASS_FAB_RECIPES, ClientHandler.MASS_FAB_OVERLAY, 3));
-        MASS_FAB[3] = create(2178, new SimpleMachineMetaTileEntity(location("mass_fab.ev"), GARecipeMaps.MASS_FAB_RECIPES, ClientHandler.MASS_FAB_OVERLAY, 4));
         if (GAConfig.GT5U.highTierMassFabs) {
```
TechLord22, line 776: UU Matter is currently required for progression. The wetware mainframe uses it in its recipe. Disabling Mass Fabricators would prevent it, so either a config warning or a removal of this config should happen.

src/main/java/gregicadditions/recipes/MatterReplication.java
```diff
@@ -2,33 +2,88 @@
...
-        for (Material m : Material.MATERIAL_REGISTRY) {
-            if (!m.hasFlag(DISABLE_REPLICATION) && m.getProtons() > 0 && m.getNeutrons() > 0 && m.getMass() != 98 && m instanceof FluidMaterial && OreDictUnifier.get(OrePrefix.dust, m).isEmpty() && m != Materials.Air && m != Materials.LiquidAir) {
-                GARecipeMaps.REPLICATOR_RECIPES.recipeBuilder().duration((int) (m.getAverageMass() * GAConfig.Misc.replicationTimeFactor * 2)).EUt(32)
-                        .inputs(FluidCellIngredient.getIngredient((FluidMaterial) m, 0))
-                        .fluidOutputs(((FluidMaterial) m).getFluid(1000)).fluidInputs(GAMaterials.PositiveMatter.getFluid((int) (m.getAverageProtons())), GAMaterials.NeutralMatter.getFluid((int) (m.getAverageNeutrons())))
+        GARecipeMaps.LARGE_MIXER_RECIPES.recipeBuilder()
+                .fluidInputs(BosionicUUMatter.getFluid(1000))
+                .fluidInputs(FermionicUUMatter.getFluid(1000))
+                .fluidInputs(FreeElectrons.getFluid(2000))
+                .fluidOutputs(Materials.UUMatter.getFluid(1000))
```
TechLord22, line 25: Technically mole standards would require this to output 4000mb of UU Matter, however I think it should probably stay at 1000mb for game balance reasons.

src/main/resources/assets/gtadditions/lang/en_us.lang
```diff
@@ -3909,6 +3909,9 @@ fluid.propadiene=Propadiene
 fluid.cyclopentadiene=Cyclopentadiene
 fluid.chloroauric_acid=Chloroauric Acid
 fluid.helium4=Helium 4
+fluid.fermionic_uu_matter=Fermionic UU-Matter
+fluid.bosionic_uu_matter=Bosionic UU-Matter
```
TechLord22, line 3913: The typo mentioned by Zalgo may apply here as well.

TechLord22 (approved):
There are no progression-related issues created by the chain, however it does make wetware mainframes harder to obtain as they use UU Matter in their recipe. I think this is okay, but it may warrant further discussion.

src/main/java/gregicadditions/recipes/MatterReplication.java
hjae78: Ah I was a bit confused about the mole standard and originally had it like that

hjae78:
We have a fluid called Free Electron Gas created by separating Helium plasma. Should Free Electrons be changed to this fluid, should Free Electrons be renamed to reduce confusion or should I leave this as is?

src/main/java/gregicadditions/machines/GATileEntities.java
hjae78: Well spotted, I think I might just remove that config option for now and reenable it when I add the multiblock. Thoughts?

Zalgo239:
> 
> 
> We have a fluid called Free Electron Gas created by separating Helium plasma. Should Free Electrons be changed to this fluid, should Free Electrons be renamed to reduce confusion or should I leave this as is?

In theory they should be the exact same thing, using a single fluid to keep things consistent makes sense.

Zalgo239 (approved):

TechLord22:
I agree with Zalgo. Additionally, we need a recipe for the separation electromagnet to make it as well.

PR #469, opened 2021-04-29
Buff Poor Ores

TechLord22:
This PR changes Poor Ores to no longer use chanced outputs on the crushed output. Currently, Poor Ores are next to useless as a result of the chanced nature, and are better furnace smelted than processed any further. This PR addresses this by changing the output type to a regular output.


hjae78:
I have no idea why I made poor ore give 1/4th the main output of normal ore...

DStrand1 (approved):

PR #471, opened 2021-05-01
Rewrite build.gradle

DStrand1:
This PR fully rewrites our build.gradle file, as it was starting to accumulate some unnecessary code and repositories. The changes I made:
- **Rewrite build.gradle in Kotlin DSL over Groovy**: Kotlin is, in my opinion, a much more capable language for this, and I have had better results using it over Groovy. With this change, doing more advanced things like CurseForge deployment, changelog generation, etc. will be easier to accomplish in the future
- **Remove Old Repositories**: I removed two repositories, being `https://maven.teamacronymcoders.com/` and `http://maven.shadowfacts.net/`. I don't think we will have any issues with these gone, but it should be something to keep an eye out for if someone has issues with it
- **Move to GTCE Fork of ForgeGradle**: We had hotfixed our environment to use my personal fork of ForgeGradle 2.3, but since then it has been put onto the [GregTechCE GitHub Organization](https://github.com/GregTechCE/ForgeGradle), which is where it will continue to receive updates instead of my personal fork
- **Extract all versioning info into `build.properties`**: I moved all versioning info for dependencies, forge, MC, JUnit, and **our own mod** into `build.properties`. Now updating dependencies is much easier and clearer, and updating our mod is as well, just changing one line at the top of the new file.
- **Move all unnecessary runtime dependencies to classpath only**: Previously, our development environment took a very long time to launch, with mods like AE2, Tinkers, RS, Forestry, etc. eating up a lot of load time. Now, I have narrowed it down to just 8 mods:
    - GregTech: CE
    - CCL
    - ChickenASM
    - JEI
    - TOP
    - CraftTweaker
    - FMP, CTM

    All other mods are still in our classpath, so there will be no compilation issues. I left comments in `build.gradle.kts` that explains how to add any one mod back to the runtime if the developer desires. I think it is best to have the absolute minimum instance on our repo, and add mods back in if we personally want them, that way everyone starts with a fast and good performance environment.


TechLord22 (approved):

DStrand1:
This should probably be merged after the 0.22 release just in case some people have issues with it

PR #477, opened 2021-05-03
Battery Tower GUI improvements, buff Plasma Turbine to using LuV Casing

hjae78:


DStrand1 (approved):
I like this PR, but if you don't mind, can I include the recipe change instead in #478? It will make dealing with git a lot easier

hjae78:
Reverted it for you

PR #478, opened 2021-05-04
Restructure Recipe Subdirectory

DStrand1:
For a long time we have had issues with files like `GARecipeAddition` being far too large to read through, `GARecipeRemoval` failing to remove recipes when they do exist, and not being able to find where things are outside of chains. This PR attempts to resolve that, as well as fixing most issues in our autogenerated recipes, errored recipe removals, and more.

## Organization
`GARecipeAddition` and `GARecipeRemoval`, as well as a handful of other dated files, are gone. Their recipe additions and removals have been reorganized, and `RecipeHandler` is now acting as our "entry point" into recipe generation for `CommonProxy`. A brief synopsis on changes:
- Major recipe additions can be put into 3 categories:
    - **Autogenerated:** Handled in `RecipeHandler` as well as a few other Handler classes in `recipes/categories/handlers`
    - **Chains:** Almost entirely unchanged, but a few recipes made their way into `VariousChains`, consisting of all the chemistry recipes that were previously in `GARecipeAddition`. It is also organized somewhat with different method calls to make it cleaner.
    - **Recipe Categories:** This is where most of the organization and work went into.

Additionally, recipe removal is no longer handled in separate classes, it is now right next to the replacement recipe. If it is a removal with no replacement, it should go into `RecipeOverride#recipeRemoval()`. Recipes should not be removed in chains, but instead either in autogenerated or in recipe categories, otherwise they could fail for being run too early.

## Recipe Categories
There are 2 subcategories in here aside from normal classes:
- **Machines:** I broke machine recipes into 3 classes, being the old `MachineCraftingRecipes` which handles hulls, casings, and some miscellaneous things like Drums, Crates, Energy Converters, etc.. Other than that, there is a class for Multiblock Recipes and Singleblock Recipes.
- **Handlers:** This is primarily a place for files I didn't need to touch, like `DisassemblyHandler`, `VoidMinerHandler`, `NuclearHandler`, etc., though I did clean up those classes a fair bit as well.

The rest of the recipes are divided into fairly straightforward classes. For example, there is a class, `ComponentRecipes`, that is for the Motor, Piston, Robot Arm, etc. recipes. There is a class called `RecipeOverride`, which handles most GTCE and Vanilla recipe changes, subdivided into sections for adding circuits, vanilla vs GT recipe removals, and just disabling recipes entirely. There are only 6 classes for this, that are subdivided further, but I think that they are clear and easy to find where any recipe is registered by following the class -> method name -> recipe comment (as I also left comments on every recipe registration).

## Recipe Changes
Now, as a result of this, I found a huge amount of issues with some parts of our recipe code dating back to Gregic Additions. Lots of unnecessary recipe removals and replacements (especially with the machines), broken recipe removals, and poorly done autogenerated recipes. A small list of changes (definitely not inclusive, as there were so many):
- Tiny/Large pipes will now have crafting recipes, with or without curved plate recipes enabled
- Small dusts no longer have 2 dust -> small dust crafting recipes
- Rings, Rounds, Foils, Double Ingots, and Curved Plates no longer generate handcrafting recipes for materials with the `NO_SMASHING` flag (mostly plastics) to be consistent with GTCE (additionally, Double Ingots will no longer generate for materials with this flag)
- Rotors will now always have an Assembler recipe (where before they only had one with Curved Plates enabled)
- Turbine Blades no longer have handcrafting recipes (fixed broken recipe removal). Additionally, they no longer take Dense Plates to make the recipe more fair
- Gem Tool heads will now use a Hammer in the recipe to craft them (this allowed me to remove a section of code that was iterating around ~70 million times on average with a minimal instance)
- Soldering alloy amounts were rebalanced somewhat for many component recipes (though more work could be done)
- Many many crafting recipe names will change. This is in part to me fixing a lot of errored recipe removals and removing redundant recipe re-registration, but also because of strange behavior with how Forge and Vanilla MC interpret the owner of GTCE recipes registered by `OrePrefix` handlers we add onto. This causes a lot of GTCE recipes to be registered under the "gtadditions" namespace (hence the "Dangerous Alternative Prefix" warnings). Since I added a few new handlers, being [pipeTiny, pipeLarge, rotor], more recipes will be under our name now

Changes in anticipation of @TechLord22's casing rework PR:
- GTCE Casings handcrafting recipes are no longer removed
- Autogenerated casing recipes have their EUt/duration set to match GTCE, and have "autogen" in their recipe name now

## Other Changes
On top of all of this, the syntax used for recipe registration was cleaned up significantly, organized better, and made more readable all around. Additionally, I reworked a lot of the autogenerated logic to be more optimal.

I also tested every single recipe removal to guarantee that:
- Recipe removals are not being called too early, before GTCE registers them
- Broken recipe removals are fixed, and
- Autogenerated recipe removal is optimized
Currently, there is still one major section of code I want to rework, being `RecipeHandler#processWireGt`, as with its current implementation, it is attempting and failing at removing thousands of recipes, if not tens of thousands at every launch. However, I wanted to open this PR now to get some feedback and make it easier for other developers to add new recipes.

I significantly cleaned up our recipe registration in `CommonProxy`, and left some relatively helpful comments to explain what is going on.

I moved all helper methods in recipe addition classes into a class called `HelperMethods`, and created many more as well. It is heavily commented, so that others can use it effectively.

In total, the changes in this PR save me 5-10 seconds on average on each launch, on top of the better organization and fixes. This will vary from person to person, though this is probably on the low-end of time savings since my average launch time is normally about 1:30.

This PR is still a Draft, so please give any suggestions on the organization to make it better if need be. However, it is nearly done, and could be merged right now if approved.

This PR should be checked out to local to review, as there are so many changes that the git diff is not very helpful.


DStrand1:
With the latest commit, this PR now includes the Plasma Turbine recipe change from #477

ALongStringOfNumbers:
In regards to the rotor assembler recipe mentioned in the opening statement, you removed this recipe in GTCE when you added the fluid solidifier recipe. Would it make sense to copy GTCE and remove both the curved and non-curved assembler recipe?

DStrand1:
With the last commit, I added a new flag for rounds, `GENERATE_ROUND`, as they were using the Small Gear flag previously. I added this flag to all materials whose rounds were used by default by Gregicality, so this means about half of the rounds from before will no longer be generated.

TO PACK MAKERS:
Use the flag `GENERATE_ROUND` to add your missing rounds back if you need them, and for any future rounds.

TechLord22 (approved):

hjae78 (approved):

PR #482, opened 2021-05-13
Enable steam mixer to output items

Puzzlebark:
Add an item export handler to the steam mixer's tile entity, as well as the appropriate UI widget. This change is intended for modpack development; it should not break the progression of stock GTCE+Gregicality, and a steam-specific recipe map should not be necessary.


DStrand1:
I cannot think of any reason why the Steam Mixer has its item output removed, so I am in favor of this change

PR #483, opened 2021-05-14
Add Tooltip Ability to Simple Fluids in CT

DStrand1:
This PR very simply allows for an `@Optional String` to be passed in the `registerFluid` CTUtils method, so that modpack makers can utilize our Fluid Tooltip system when adding new Simple Fluids. Since it is marked as `optional`, it will not break any existing scripts, and modpack authors simply have to add a String after the name and rgb values where they make fluids already.


rhgngle:
What about simpledust? I'd love to see one for that as well

PR #488, opened 2021-05-17
Component Recipe Cleanup

DStrand1:
This PR addresses a few issues remaining from the cleanup PR:

- I forgot to register the GENERATE_ROUND material flag, so I went ahead and did that
- Added a couple tooltip fixes to known issues
- Fixed a recipe removal for Fiber Reinforced Epoxy Dust causing the small and tiny dust recipes to disappear
- Removed the Steel Steam Furnace recipe override since the issue was fixed in GTCE: https://github.com/GregTechCE/GregTech/pull/1587

Lastly and most importantly, I did a final take on the component assembler recipes.
First off, I cut the times down to be 5s with on-tier voltage. I changed the voltages to be the "standard" voltages, being 30, 120, 480, 1920, 7680. Next, I cut out some of the more annoying pieces of the assembler component recipes (emitters/sensors taking flawless gems, pumps taking soldering alloy and rings, pistons taking curved plates, etc.). All in all, these recipes are only slightly cheaper, and in some ways may be more difficult, and of course take power. However, I think they are now a viable alternative to handcrafting recipes for automation considering they are on average cheaper.

Lastly I made the LuV-UEV component recipes consistent with Soldering Alloy, being `L, L, L, L*2, L*4` for the tiers respectively. Also, LuV+ Sensors and Emitters were using the wrong marker material for their tiered circuits, so they are now using on-tier circuits.


src/main/java/gregicadditions/recipes/categories/ComponentRecipes.java
```diff
@@ -193,7 +198,7 @@ private static void emitterInit() {
                 .input(foil, Electrum, 64)
                 .input(wireGtDouble, YttriumBariumCuprate, 8)
                 .input(gemExquisite, Ruby, 2)
-                .input(circuit, Extreme, 2)
+                .input(circuit, Master, 2)
```
pablocnvs, line 201: Didn't we lower the circuit tier because of the Supercausal Mainframe thing

TechLord22 (approved):

ALongStringOfNumbers (approved):
Looks good to me as is. One thing to consider is if discounts for the pump assembler recipes fluid inputs want to be given for using more advanced materials.

PR #492, opened 2021-05-20
Steam chunk miner, chunk miner overlays, steam machine tweaks

Puzzlebark:
This PR contains 4 changes, 3 of which have to do with steam machines and 1 cosmetic addition:
- A new steam chunk miner that performs one mining operation every 8 ticks and requires 16mb/t of steam. The expensive recipe as well as the setup required to get good use out of it are the tradeoff for having early auto-mining, but it is subject to change with feedback.

image

- The steam mixer's input slots can now transfer fluid to and from containers.

- The steam macerator recipes have been reverted to using diamonds instead of flint.

- Electric chunk miners have new overlay textures on their horizontal faces. 

image


src/main/java/gregicadditions/machines/multi/miner/Miner.java
```diff
@@ -18,6 +18,7 @@
 public interface Miner {
 
     enum Type {
+        STEAM(8, 1, 0, "", 1),
```
DStrand1, line 21: This is probably fine, but it is bad practice to insert values into an Enum anywhere but the end, as this changes their ordinals. I looked and saw no places where this is a problem, but if you could sanity check this and make sure that `ordinal()` or `values()` is not called on this Enum, that would be great.

src/main/java/gregicadditions/machines/multi/miner/TileEntitySteamMiner.java
```diff
@@ -0,0 +1,227 @@
...
+import net.minecraft.nbt.NBTTagLong;
+import net.minecraft.util.*;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.util.math.ChunkPos;
+import net.minecraft.world.World;
+import net.minecraft.world.WorldServer;
+import net.minecraft.world.chunk.Chunk;
+import net.minecraftforge.fluids.FluidActionResult;
+import net.minecraftforge.fluids.FluidStack;
+import net.minecraftforge.fluids.FluidTank;
```
DStrand1, line 40: There are some extraneous imports in here

src/main/java/gregicadditions/machines/multi/miner/TileEntitySteamMiner.java
```diff
@@ -0,0 +1,227 @@
...
+import java.util.List;
+import java.util.Objects;
+import java.util.concurrent.atomic.AtomicLong;
+
+public class TileEntitySteamMiner extends MetaTileEntity implements Miner {
+
+    private final int inventorySize;
+    public final Miner.Type minerType;
+    private AtomicLong x = new AtomicLong(Long.MAX_VALUE), y = new AtomicLong(Long.MAX_VALUE), z = new AtomicLong(Long.MAX_VALUE);
+    private final ItemStackHandler fluidContainerInventory;
```
DStrand1, line 63: This needs to be saved in `writeNBT` and `readNBT` or else items in the Fluid Container slots will be voided when the machine is unloaded

src/main/java/gregicadditions/machines/multi/miner/TileEntitySteamMiner.java
```diff
@@ -0,0 +1,227 @@
...
+    public void update() {
+        super.update();
+        if (!getWorld().isRemote) {
+            fillInternalTankFromFluidContainer(fluidContainerInventory, fluidContainerInventory, 0, 1);
+            // if sufficient amounts of steam and drilling fluid aren't present, do nothing
+            if ((importFluids.getTankAt(0).getFluidAmount() < 1) || (importFluids.getTankAt(1).getFluidAmount() < 16)) {
+                return;
+            } else {
+                System.out.println("im draining brooooooo");
+            }
```
DStrand1, line 165:
```suggestion
            if ((importFluids.getTankAt(0).getFluidAmount() < 1) || (importFluids.getTankAt(1).getFluidAmount() < 16)) {
                return;
            }
```

src/main/java/gregicadditions/recipes/categories/machines/SingleblockCraftingRecipes.java
```diff
@@ -344,8 +344,8 @@ private static void singleblockOverride() {
...
-        ModHandler.addShapedRecipe("ga_steam_macerator_steel",  MetaTileEntities.STEAM_MACERATOR_STEEL.getStackForm(),  "DXD", "XMX", "PXP", 'M', MetaBlocks.MACHINE_CASING.getItemVariant(BlockMachineCasing.MachineCasingType.STEEL_HULL),        'X', new UnificationEntry(pipeSmall, Steel),  'P', OreDictNames.craftingPiston, 'D', new ItemStack(Items.FLINT));
+        ModHandler.addShapedRecipe("ga_steam_macerator_bronze", MetaTileEntities.STEAM_MACERATOR_BRONZE.getStackForm(), "DXD", "XMX", "PXP", 'M', MetaBlocks.MACHINE_CASING.getItemVariant(BlockMachineCasing.MachineCasingType.BRONZE_HULL),       'X', new UnificationEntry(pipeSmall, Bronze), 'P', OreDictNames.craftingPiston, 'D', new ItemStack(Items.DIAMOND));
+        ModHandler.addShapedRecipe("ga_steam_macerator_steel",  MetaTileEntities.STEAM_MACERATOR_STEEL.getStackForm(),  "DXD", "XMX", "PXP", 'M', MetaBlocks.MACHINE_CASING.getItemVariant(BlockMachineCasing.MachineCasingType.STEEL_HULL),        'X', new UnificationEntry(pipeSmall, Steel),  'P', OreDictNames.craftingPiston, 'D', new ItemStack(Items.DIAMOND));
```
DStrand1, line 348: Instead of updating these, you can simply remove the recipe removals and additions

src/main/java/gregicadditions/machines/multi/miner/TileEntitySteamMiner.java
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    @Override
+    public void readFromNBT(NBTTagCompound data) {
+        super.readFromNBT(data);
+        x.set(data.getLong("xPos"));
+        y.set(data.getLong("yPos"));
+        z.set(data.getLong("zPos"));
+    }
+
```
DStrand1, line 222: The items in the Fluid Container slots are voided if the machine is broken. This section of code will fix that (you will have to import `net.minecraft.block.Block`):
```suggestion

        @Override
        public void onRemoval() {
            for (int slot = 0; slot < fluidContainerInventory.getSlots(); slot++)
                Block.spawnAsEntity(getWorld(), getPos(), fluidContainerInventory.getStackInSlot(slot));
        }

```

DStrand1 (changes requested):
The functionality of this is great! I'm surprised at how well you were able to get the Fluid Handler item slot to work, properly delegating Steam and Drilling Fluid into the right internal fluid tanks. There are just a few places in the code I would like to see cleaned up

DStrand1:
Also I asked for the Steam Macerator changes to be in this PR, and @TechLord22 asked for the Steam Mixer fluid slot changes, so that is why 2 seemingly unrelated changes are here.

src/main/java/gregicadditions/machines/multi/miner/Miner.java
Puzzlebark: I'll keep this in mind in the future. Neither of those methods seem to be called on this Enum.

src/main/java/gregicadditions/machines/multi/miner/TileEntitySteamMiner.java
Puzzlebark: Got it, the redundant imports have been removed.

src/main/java/gregicadditions/machines/multi/miner/TileEntitySteamMiner.java
Puzzlebark: My bad, forgot to remove that before opening the PR haha

src/main/java/gregicadditions/machines/multi/miner/TileEntitySteamMiner.java
Puzzlebark: Good catch, that's been added now.

src/main/java/gregicadditions/machines/multi/miner/TileEntitySteamMiner.java
Puzzlebark: Got that implemented.

DStrand1 (approved):

DStrand1:
In further Discord discussion, we realized that this machine is not consistent with other Steam machines with its output slot. It should vent Steam for each broken block, and not have item output capabilities. Additionally if it is unable to vent, the machine should stop like other Steam machines

Puzzlebark:
The output face has now been replaced with an exhaust face.

DStrand1:
Seems there is a lot of older commits coming along with this PR. If you do an interactive rebase, you can clean that up.

`git rebase -i HEAD~<some number of commits>`
for example, `git rebase -i HEAD~20`. Then you will be able to pick and choose which commits you want to keep

Puzzlebark:
Fixed.

Puzzlebark:
The steam miner no longer responds to right-clicking with a wrench. All the issues with the steam miner I'm aware of have been addressed.

TechLord22 (approved):

PR #497, opened 2021-05-20
Add Double Plates and Delete Double Ingots

bruberu:
This particular PR does as the title says.
This does make a config option obsolete, and potentially deletes double ingots from worlds, although I must say that I don't believe that most people just keep them around.


src/main/java/gregicadditions/GAMaterials.java
```diff
@@ -42,17 +42,21 @@
...
         Material.MatFlags.registerMaterialFlag("GENERATE_ROUND", GENERATE_ROUND, IngotMaterial.class);
+        Material.MatFlags.registerMaterialFlag("GENERATE_DOUBLE_PLATE", GENERATE_DOUBLE_PLATE, IngotMaterial.class);
+
     }
 
     public static long STD_METAL = GENERATE_PLATE;
-    static long EXT2_METAL = GENERATE_PLATE | GENERATE_DENSE | GENERATE_ROD | GENERATE_BOLT_SCREW | GENERATE_GEAR | GENERATE_FOIL | GENERATE_FINE_WIRE | GENERATE_LONG_ROD;
-    static long CORE_METAL = EXT2_METAL | GENERATE_RING | GENERATE_FRAME | GENERATE_ROTOR | GENERATE_SMALL_GEAR | GENERATE_DENSE;
+    static long EXT2_METAL = GENERATE_PLATE | GENERATE_DENSE | GENERATE_ROD | GENERATE_BOLT_SCREW | GENERATE_GEAR | GENERATE_FOIL | GENERATE_FINE_WIRE | GENERATE_LONG_ROD | GENERATE_DOUBLE_PLATE;
+    static long CORE_METAL = EXT2_METAL | GENERATE_RING | GENERATE_FRAME | GENERATE_ROTOR | GENERATE_SMALL_GEAR | GENERATE_DENSE | GENERATE_DOUBLE_PLATE;
```
DStrand1, line 59: The flag only needs to be applied to EXT2_METAL since CORE_METAL already contains all the flags of EXT2_METAL
```suggestion
    static long CORE_METAL = EXT2_METAL | GENERATE_RING | GENERATE_FRAME | GENERATE_ROTOR | GENERATE_SMALL_GEAR | GENERATE_DENSE;
```

src/main/java/gregicadditions/item/GAMetaItem.java
```diff
@@ -30,14 +30,16 @@
 public class GAMetaItem extends MaterialMetaItem {
     public GAMetaItem() {
         super(GAConfig.GT6.addCurvedPlates ? GAEnums.GAOrePrefix.plateCurved : null,
-                GAConfig.GT6.addDoubleIngots ? GAEnums.GAOrePrefix.ingotDouble : null,
```
DStrand1, line 33: We may as well put the double plate prefix here so that if players do happen to have a double ingot for some material, it will get converted to a double plate

DStrand1 (changes requested):
Just a couple small things, otherwise I think this PR is great! Will really help balance out  turbine blades and battery tower cells, and other things in the future

DStrand1 (approved):

TechLord22 (approved):
Nice work!

bruberu:
Note: I fixed the bug with compressor recipes registering for the wrong items (e.g. chests) within the latest release, as I don't really know how to manage multiple PRs yet, but it should still be a useful thing to have here.

src/main/java/gregicadditions/recipes/categories/CasingRecipes.java
```diff
@@ -180,46 +181,60 @@ private static void tieredCasings() {
...
+                .inputs(GAMetaBlocks.CELL_CASING.getItemVariant(CELL_UV))
+                .outputs(GAMetaBlocks.CELL_CASING.getItemVariant(CELL_UHV))
+                .buildAndRegister();
+
+        ASSEMBLER_RECIPES.recipeBuilder().EUt(3932160).duration(1200)
+                .input(plateDouble, Bohrium, 4)
+                .fluidInputs(FreeAlphaGas.getFluid(4000))
+                .inputs(GAMetaBlocks.CELL_CASING.getItemVariant(CELL_UHV))
+                .outputs(GAMetaBlocks.CELL_CASING.getItemVariant(CELL_UEV))
+                .buildAndRegister();
```
hjae78, line 237: I'm a little hesitant about buffing this recipe so much. Compared to the corresponding batteries the cells are much cheaper

bruberu: Considering that making these batteries functional requires one making at least 36 of these, I personally believe that this would make the recipes considerably less painful to deal with in batches. Besides, some of the gases are already not terribly easy to get in bulk, so this should be balanced enough.

PR #499, opened 2021-05-22
Fix Nuclear Fluorine

TechLord22:
This PR addresses the massive loss in fluorine resulting from the Nuclear Process. 

First, this PR fixes the remaining chemistry problems within the NuclearHandler, except for one recipe which would require extra item inputs and would cause significant problems for existing setups. The Nitrite materials are also actually supposed to be Nitrates, and the chemistry reflects this. Durations and EUts have been made static for all inputs, allowing automation of the process to be more standardized. The durations have also been reduced.

The Gas Centrifuge is the cause of the fluorine loss, and could also lead to positive fluorine loops. This PR changes the gas centrifuge to only have one recipe per nuclear input fluid, which is equivalent to the old ULV-tier recipe. This recipe uses 2A of HV (960EUt) and retains the same duration logic currently used. The recipe has been changed to no longer use an integrated circuit so existing setups will automatically start using the new recipe. No changes to the multiblock structure are required.

The fluorine input and output is now mathematically identical, and one can expect to recover ~118/120B with ease. The other two buckets will take longer, as they are output in smaller amounts and need 1000mB input to be recovered. 

As a result of the old ULV-tier recipe becoming the sole one, some machines are much harder to create (such as the MV decay chamber) due to the reduced isotope output for the less common variants. The most affected recipes have been altered to use the most common isotope in order to offset this.

### DIFF
--- a/src/main/java/gregicadditions/recipes/categories/handlers/NuclearHandler.java
+++ b/src/main/java/gregicadditions/recipes/categories/handlers/NuclearHandler.java
@@ -233,8 +233,8 @@ public class NuclearHandler {
             radioactiveMaterial.composition.forEach((key, value) -> builder1.fluidOutputs(key.getFluidHexafluoride(value * 2)));
             builder1.buildAndRegister();
 
-            // [Mat + F6] + 2H2O -> [Mat + F6 + 2H2O]
             radioactiveMaterial.composition.forEach((key, value) -> {
+                // [Mat + F6] + 2H2O -> [Mat + F6 + 2H2O]
                 CRACKING_RECIPES.recipeBuilder().duration(40 * complexity / 100).EUt(120)
                         .fluidInputs(Steam.getFluid(2000))
                         .fluidInputs(key.getFluidHexafluoride(1000))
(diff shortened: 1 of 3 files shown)


DStrand1 (approved):
This will be nice to not lose lots of Fluorine until we can do a proper nuclear rework

src/main/java/gregicadditions/recipes/categories/handlers/NuclearHandler.java
```diff
@@ -198,108 +198,60 @@ private static void processNuclearMaterial(OrePrefix ingot, IngotMaterial materi
...
-                        .fluidOutputs(key.getFluidHexafluorideSteamCracked(7000))
+                // [Mat + F6] + 2H2O -> [Mat + F6 + 2H2O]
+                CRACKING_RECIPES.recipeBuilder().duration(40 * complexity / 100).EUt(120)
+                        .fluidInputs(Steam.getFluid(2000))
+                        .fluidInputs(key.getFluidHexafluoride(1000))
+                        .fluidOutputs(key.getFluidHexafluorideSteamCracked(1000))
                         .buildAndRegister();
-                BLAST_RECIPES.recipeBuilder().blastFurnaceTemp(600).duration(600 * complexity / 100).EUt(120 * complexity / 100)
+
+                // [Mat + F6 + 2H2O] -> [Mat + 2O] + 6F + 4H (H lost)
```
pablocnvs, line 244: You could make the cracker take 3000mb of steam and then output in this step 6000mb of hydrofluoric acid

ALongStringOfNumbers:
This PR needs to be fixed, as some unrelated commits have snuck in.

PR #503, opened 2021-05-23
Rebalance Energy Usage and Duration of Recipes, Add Combined Processes

TechLord22:
This PR primarily does two things. First, it changes the energy consumption and usage of almost all of our recipes. Second, it combines so major processes into late-game chemical plant recipes. Additionally, some minor chemistry issues have been fixed.

The energy costs have changed as follows: early-chains can expect to use LV-MV power, mid-ish-game chains can expect HV-EV, later-game chains use IV-LuV, and even later chains will use ZPM-UEV power, depending on the individual chain and recipe. Durations have mostly been decreased overall, in order to incentivize the use of our chains more. Overall, one can expect faster processing at less power for the majority of recipes.

For the combined processes, the general rule for energy usage is 2 voltage tiers more than the earliest tier the material is used at. Durations have been kept quite short as a result of the increased energy usage. Some materials, such as PVC, did not get a combined process because it is too short to warrant it. Naquadah also received a combined chain due to its late-game usage, and the chemistry works out such that the Naquadah Compound turns directly to the regular dust without any extra inputs or outputs. 

I have also added an additional and optional way to process sulfuric oil fractions (naphtha, light fuel, etc.) using Hydrodesulfurization catalysts. These recipes are functionally identical to the regular recipes, except that they process half the input at twice the speed. This leaves the overall fuel rate the same, but doubles the sulfur gain. That is the purpose of these cataysts - extracting more sulfur from these fractions.

Some bad GTCE recipes have also been fixed, and potential positive power loops from Hydrogen in the gas turbine have been prevented by making Hydrogen a useless fuel for all intents and purposes.

Thank you to @DStrand1 for fixing the commit problems.


DStrand1:
First force-push was to fix git hierarchy issues. Second was to rebase against the commits for Gregicality 0.22.4

src/main/java/gregicadditions/recipes/chain/CombinedChains.java
```diff
@@ -0,0 +1,158 @@
...
+        CHEMICAL_PLANT_RECIPES.recipeBuilder().duration(50).EUt(491520) // UV
+                .input(dust, EnrichedNaquadricCompound)
+                .output(dust, NaquadahEnriched)
+                .buildAndRegister();
+
+        // Naquadriatic Compound [*Nq*] -> *Nq*
+        CHEMICAL_PLANT_RECIPES.recipeBuilder().duration(50).EUt(491520) // UV
+                .input(dust, NaquadriaticCompound)
+                .output(dust, Naquadria)
+                .buildAndRegister();
```
hjae78, line 40: I'm not sure about these recipes, would prefer to see the same inputs here

TechLord22: I changed up these recipes to now be more reflective/reminiscent of the unique portion of each naquadah variant's processing.

PR #508, opened 2021-05-25
Rework Broken Configs from GA

DStrand1:
There are a lot of relatively minor changes in here, so let me just list them all out:
- I compacted and separated all the Bending and Curved Plate configs into just two:
    - BendingCylinders
    - addCurvedPlates
They are no longer coupled to eachother in any way, so Curved Plates have recipes without bending cylinders, and bending cylinders have uses without curved plates.
- The config option `BendingFoilsAutomatic` will now disable all Cluster Mill machines if false. This is a change I would like discussion for because I'm not sure it is a good idea, but I made it to be consistent with other config options we have (like Disassembler)
- I moved the `registerCrates` config option out of `Misc` and into `GT6` to be next to the Drums config option
- I removed the `GeneratedSawingRecipes` config, since it was fully unused, and as far as I could tell, fully implemented as a GTCE config
- I removed the `PackagerDustRecipes` config value. It was only being used to add a schematic to Small Dust packaging recipes, when this schematic was always being added to other dusts anyway
- I removed the `Unpackager3x3Recipes` config and combined it with the `RemoveBlockUncraftingRecipes` config, Now if this is enabled, hand uncrafting recipes will be removed and added to the unpackager instead
- I removed the `Packager3x3Recipes` config, and combined it with the `Remove3x3BlockRecipes` config. Now if this is enabled, handcrafting recipes will be removed, and packager recipes registered in their place
- I moved the `Packager2x2Recipes` config to be with the others in the `GT5U` section

As a result of all of this, ***config files will need to be regenerated upon updating to the new version of Gregicality***.

Additionally, I reworked the `RecipeHandler#generatedRecipes()` further to clean it up and fix more issues with it:
- I removed the Forge Hammer recipe generation from it
- Nuggets will now be removed from handcrafting when before they were exempt
- Ingots can now be Packaged/Unpackaged/Compressed where previously they would cause it to return

All of these changes are in the effort of making the config description completely accurate. Before they were confusing both to the player, and clearly to the original author, since they were poorly used and inconsistently used.

Lastly, I made a few small recipe changes while I was changing things:
- Buffed Schematics to use LV circuits and Steel instead of MV circuits and Stainless steel. Also requires less of both. This was done since schematics are more forced now following some of these config changes
- Fix a couple Gold sources and replace with Precious Metal: Lava, Netherrack Maceration, Glowstone
- Add a Glowstone Mixer recipe to make the dust:
    - 1 Redstone + 1 Precious Metal = 1 Glowstone
    - 1 Redstone + 1 Gold = 2 Glowstone

Additionally, there was an issue where both our `GENERATE_CASING` flag and GTCE's `EXCLUDE_BLOCK_CRAFTING_BY_HAND_RECIPES` flags were assigned the same IDs, being 47. This led to all of our casing blocks having no handcrafting block recipes, leading to a few bugs mentioned regularly in Discord. This issue took me hours to find, but now will no longer be an issue.

Since the flag worked both ways, there were some materials that were given frames and casings when we didn't want them to. The only one I could find was Gold, so Gold no longer has a frame or casing now.


src/main/java/gregicadditions/recipes/RecipeHandler.java
```diff
@@ -1414,124 +1418,119 @@ private static void matterReplication(Material material) {
...
-                                .buildAndRegister();
-                    }
-                }
-            }
+            removeRecipeByName(recipe.getRegistryName());
+            PACKER_RECIPES.recipeBuilder().duration(100).EUt(4)
+                    .inputs(CountableIngredient.from(input, 9))
+                    .notConsumable(SCHEMATIC_3X3.getStackForm())
+                    .outputs(output)
+                    .buildAndRegister();
```
TechLord22, line 1465: I'd like to see this recipe always registered instead of only with the config. The config instead should solely decide whether to remove them. If schematics are not to be used (as a result of other configs), a circuit of 3 should be used instead.

src/main/java/gregicadditions/recipes/RecipeHandler.java
```diff
@@ -1414,124 +1418,119 @@ private static void matterReplication(Material material) {
...
+                return;
 
-                    if (!recipesToRemove.contains(recipe.getRegistryName())
-                            && !hasPrefix(recipe.getRecipeOutput(), "dust", "dustSmall")
-                            && recipe.getRecipeOutput().getCount() == 1
-                            && GAConfig.Misc.Packager2x2Recipes) {
+            COMPRESSOR_RECIPES.recipeBuilder().duration(400).EUt(2)
+                    .inputs(CountableIngredient.from(input, 9))
+                    .outputs(output)
+                    .buildAndRegister();
```
TechLord22, line 1478: I'd also like to see this recipe always registered instead of only with the config. The config instead should solely decide whether to remove them or not.

src/main/java/gregicadditions/recipes/RecipeHandler.java
```diff
@@ -1414,124 +1418,119 @@ private static void matterReplication(Material material) {
...
+        if (GAConfig.GT5U.Packager2x2Recipes) {
 
-                        // Remove Block Uncrafting Recipes
-                        if (GAConfig.GT5U.RemoveBlockUncraftingRecipes)
-                            recipesToRemove.add(recipe.getRegistryName());
+            PACKER_RECIPES.recipeBuilder().duration(100).EUt(4)
+                    .inputs(CountableIngredient.from(input, 4))
+                    .notConsumable(SCHEMATIC_2X2.getStackForm())
+                    .outputs(output)
+                    .buildAndRegister();
```
TechLord22, line 1504: Once again - I'd like to see this recipe always registered instead of only with the config. The config instead should solely decide whether to remove them. If schematics are not to be used (as a result of other configs), a circuit of 2 should be used instead.

src/main/java/gregicadditions/recipes/RecipeHandler.java
```diff
@@ -1414,124 +1418,119 @@ private static void matterReplication(Material material) {
...
-                                .buildAndRegister();
-                    }
-                }
-            }
+            removeRecipeByName(recipe.getRegistryName());
+            UNPACKER_RECIPES.recipeBuilder().duration(100).EUt(8)
+                    .inputs(input)
+                    .notConsumable(SCHEMATIC_3X3.getStackForm())
+                    .outputs(output)
+                    .buildAndRegister();
```
TechLord22, line 1533: Last time: I'd like to see this recipe always registered instead of only with the config. The config instead should solely decide whether to remove them. If schematics are not to be used (as a result of other configs), a circuit of 3 should be used instead.

src/main/java/gregicadditions/GAConfig.java
```diff
@@ -54,17 +47,17 @@
         @Config.Name("Should Drums be registered?")
         public boolean registerDrums = true;
 
+        @Config.Comment("Set this to false to disable Crates")
+        @Config.Name("Should Crates be registered?")
+        public boolean registerCrates = true;
+
```
TechLord22, line 53: I don't think these need configs. If you don't want them - removeAndHide them with CT.

src/main/java/gregicadditions/recipes/RecipeHandler.java
DStrand1: im not really sure what the gain from this is. Currently we have configs to decide to add them or not, and they default to adding them. So disabling this config removes the recipes. Additionally, schematics are not configurable, and as a result I made them significantly cheaper

src/main/java/gregicadditions/GAConfig.java
DStrand1: Seems like a reasonable change to me

src/main/java/gregicadditions/recipes/RecipeHandler.java
```diff
@@ -1414,124 +1418,121 @@ private static void matterReplication(Material material) {
...
+            PACKER_RECIPES.recipeBuilder().duration(100).EUt(4)
+                    .inputs(CountableIngredient.from(input, 9))
+                    .notConsumable(SCHEMATIC_3X3.getStackForm())
+                    .outputs(output)
+                    .buildAndRegister();
+        }
 
-            if (recipe.getIngredients().size() == 4) {
+        // Add Compressor 3x3 Recipes
+        if (GAConfig.GT5U.GenerateCompressorRecipes) {
```
ALongStringOfNumbers, line 1469: Would it be prudent to check if the recipe is already a compressor recipe added by GTCE?

src/main/java/gregicadditions/recipes/RecipeHandler.java
```diff
@@ -1414,124 +1418,121 @@ private static void matterReplication(Material material) {
...
+            return;
 
-                    if (outputIsNot(recipe, Blocks.AIR, Blocks.SLIME_BLOCK)) {
+        // Add Packager 2x2 Recipes
+        if (GAConfig.GT5U.Packager2x2Recipes) {
 
-                        // Remove Block Uncrafting Recipes
-                        if (GAConfig.GT5U.RemoveBlockUncraftingRecipes)
-                            recipesToRemove.add(recipe.getRegistryName());
+            PACKER_RECIPES.recipeBuilder().duration(100).EUt(4)
```
ALongStringOfNumbers, line 1502: Should this recipe be removed from the crafting table, like in the 3x3 case? Or is it just meant to be moved into the packager?

ALongStringOfNumbers (approved):
Generally looks good to me, left a couple comments for consideration

hjae78 (approved):

PR #519, opened 2021-06-04
updated digital interface to support larger amount of item/fluid/energy

galyfray:
One of my friend pointed out that when using the mulktiblock battery the number showed by the digital interface may  overflow from where it should be:

i fixed this by changing a little the computation in order to make easier to add new  unit tier (yeah not sure about how to says this) and adding the Giga Terra and Peta prefix hopping it should be enough

### DIFF
--- a/src/main/java/gregicadditions/covers/CoverDigitalInterface.java
+++ b/src/main/java/gregicadditions/covers/CoverDigitalInterface.java
@@ -996,7 +996,7 @@ public class CoverDigitalInterface extends CoverBehavior implements IRenderMetaT
         }
         int i = 1;
 
-        while (number / 10000000 != 0) {
+        while (number / 10000000 != 0 && i < units.length) {
             number = number / 1000;
             ++i;
         }
--- a/src/main/java/gregicadditions/covers/CoverDigitalInterface.java
+++ b/src/main/java/gregicadditions/covers/CoverDigitalInterface.java
@@ -110,6 +110,7 @@ public class CoverDigitalInterface extends CoverBehavior implements IRenderMetaT
         MACHINE,
         PROXY;
         public static MODE[] VALUES;
+
         static {
             VALUES = MODE.values();
         }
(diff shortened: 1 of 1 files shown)


galyfray:
ohhhh no I didn't pay attention about it but idea has reformated everything what a hell hope it doesn't annoy you too much I'm so sorry about that

DStrand1 (approved):
Seems like a good fix. With regards to the reformatting, it looks like this file really needed it, so no problem.

Yefancy (approved):

src/main/java/gregicadditions/covers/CoverDigitalInterface.java
```diff
@@ -962,30 +967,40 @@ private void renderItemMode(int slot) {
...
         }
-        return new DecimalFormat("#.#").format(number * 1.0f / 1000000) + units[2][unit];
+        int i = 1;
+
+        while (number / 10000000 != 0 && i < units.length) {
+            number = number / 1000;
+            ++i;
+        }
+
+        return new DecimalFormat("#.#").format(number * 1.0f / 1000) + units[i][unit];
```
Yefancy, line 1004: I think 10,000,000 is wrong here?
E.g. After calculation, 1,000,000 (eu) => 1,000.0KEU, but the expected result should be 1.0MEU.

Yefancy: If you purpose to keep a little more length of the text, for the sake of looking good, I think it's OK

galyfray: yeah I think that 1000KEU can be a bit nicer than directly going to MEU even if that wasn't planned if nobody disagree with this then let's go with this else I will remove this 0.

Yefancy:
I think it could be merged now.  And i have a pr to do recently that I could handle the conflicts with this file in advance

hjae78 (approved):

PR #522, opened 2021-06-06
config for small gears haters

### LINKED ISSUE #521: make the small gear craft a config
not everyone likes the GT5U way of doing small gears I think it should be a configurable thing such as the wrench craft changes.

galyfray:
closes #521 

This will help people who have made some automation using the older recipes.

I've once again forget about my macro override for saving and auto-formating sorry ...

 it's my first time dealing with config so I'm quite unsure if I've done the things right


DStrand1:
Since this PR had a large amount of code changes, I decided to include the changes into #523. I also wasn't in favor of all of the changes made here, so I simplified it to have a config to restore the original GTCE manual Small Gear recipes. Forge Hammer recipes will still always be replaced by Extruder Recipes.

galyfray:
The main Idea of this pr was to allow people to re enable the forge hammer recipes for people who already have done some automation using the old recipe . In addition to this you can't automate electric piston at the LV tier you have to use extruder I think many people won't see as a cool feature of disallowing early automation while playing a mod about automation ...

DStrand1:
To me, it doesn't really make sense that a small gear be made by smashing a piece of metal with a hammer, as that really doesn't seem like it would make something delicate like a gear. In any case, there is always fluid solidification, which is doable at LV and is often times still faster than extrusion

galyfray:
>To me, it doesn't really make sense that a small gear be made by smashing a piece of metal with a hammer, as that really doesn't seem like it would make something delicate like a gear. In any case, there is always fluid solidification, which is doable at LV and is often times still faster than extrusion

I agree on the forge hammer part concerning the fluid solidification if my memory is good many recipes use more than LV.

An assembler may do the job using a plate and a rod if the harder manual is true and only a plate if easier craft is activated keeping all small gear doable at LV

DStrand1:
According to the [Small Gear Handler](https://github.com/GregTechCE/GregTech/blob/dde025306a81abf2d7ad10561fcd3553805ca17e/src/main/java/gregtech/loaders/oreprocessing/PartsRecipeHandler.java#L154) in GTCE, they are all hard-coded to use 8EU/t, so they are all doable at LV.

galyfray:
I didn't express myself correctly sorry . I was meaning the fluid extraction is not at LV this one is not hard coded and is one of the reason why I made this pr.

DStrand1:
You're correct that some materials are locked to higher tiers for fluid extraction, such as Steel which is the only really useful Small Gear for LV. I'm thinking of instead just dropping Steel fluid extraction to LV, that should allow for it to be automated just fine for LV Pistons

galyfray:
I think that steel fluid at LV is okay there is still the aluminium that was useful in order to automate LV parts before getting into the tier. 

I understand that the forge hammer recipes aren't realistic but I think that an LV recipe for all small gears should exist even if it cost more than the extruder one. There is plenty of machines that can be used to create small gear. Maybe the forming press or the assembler could do the job.

DStrand1:
I've decided to implement this similarly to normal Gears, and have an Alloy Smelter recipe using the Small Gear Mold, taking 2 of the material rather than 1 like the Extruder, but at a constant 30 EU/t

galyfray:
Thank you.

PR #524, opened 2021-06-07
Update and add surface rocks to veins

ALongStringOfNumbers:
Updates some surface rock definitions to the newer surface rock format, and adds surface rocks to veins that did not have them before.

Let me know if there is any material that should be changed, as each surface rock yields a couple tiny dusts of material, and I don't know if that would mess with anything that is currently going on.

### DIFF
--- a/src/main/resources/assets/gregtech/worldgen/end/naquadah_vein2.json
+++ b/src/main/resources/assets/gregtech/worldgen/end/naquadah_vein2.json
@@ -6,7 +6,7 @@
   "dimension_filter": ["name:the_end"],
   "vein_populator": {
     "type": "surface_rock",
-    "material": "naquadah"
+    "material": "naquadric_compound"
   },
   "generator": {
     "type": "ellipsoid",
--- a/src/main/resources/assets/gregtech/worldgen/end/naquadah_vein2.json
+++ b/src/main/resources/assets/gregtech/worldgen/end/naquadah_vein2.json
@@ -4,8 +4,10 @@
   "max_height": 65,
   "min_height": 30,
   "dimension_filter": ["name:the_end"],
-  "surface_stone_material": "naquadah",
-  
+  "vein_populator": {
+    "type": "surface_rock",
+    "material": "naquadah"
+  },
   "generator": {
     "type": "ellipsoid",
     "radius": [25, 25]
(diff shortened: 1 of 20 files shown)


src/main/resources/assets/gregtech/worldgen/end/naquadah_vein2.json
```diff
@@ -4,8 +4,10 @@
   "max_height": 65,
   "min_height": 30,
   "dimension_filter": ["name:the_end"],
-  "surface_stone_material": "naquadah",
-  
+  "vein_populator": {
+    "type": "surface_rock",
+    "material": "naquadah"
```
TechLord22, line 9: Naquadah veins should definitely use the compound variants. We currently have a hard barrier of a 2A IV/1A LuV blast furnace as the gate, and I think this should stay.

src/main/resources/assets/gregtech/worldgen/overworld/platinum_vein2.json
```diff
@@ -3,8 +3,10 @@
   "density": 0.6,
   "max_height": 50,
   "min_height": 20,
-  "surface_stone_material": "platinum",
-
+  "vein_populator": {
+    "type": "surface_rock",
+    "material": "platinum"
```
TechLord22, line 8: I think only this platinum vein should be changed to platinum metallic powder, so that the overworld variant doesn't give you platinum immediately, but the end variant (which takes more "work" to obtain) does give access to small amounts.

src/main/resources/assets/gregtech/worldgen/overworld/pitchblende_vein3.json
```diff
@@ -3,7 +3,10 @@
   "density": 0.5,
   "max_height": 40,
   "min_height": 10,
-  "surface_stone_material": "uranium_radioactive",
+  "vein_populator": {
+    "type": "surface_rock",
+    "material": "uranium_radioactive"
```
warjort, line 8: This should probably be pitchblende?

TechLord22: I think regular uranium is fine here. The pitchblende/Uraninite process is really long, and a tiny amount of starter uranium isn't really a problem because of how hard the process really is.

TechLord22 (approved):

PR #525, opened 2021-06-08
Exotic Circuits

Zalgo239:
+Added Exotic circuits
+Added crafting chains for Exotic circuits
+Added localization for Exotic circuits


src/main/java/gregicadditions/recipes/chain/OrganometallicChains.java
```diff
@@ -0,0 +1,58 @@
...
+                .fluidOutputs(Dimethylether.getFluid(1000))
+                .buildAndRegister();
+
+        CHEMICAL_RECIPES.recipeBuilder().duration(160).EUt(2000)
+                .fluidInputs(Dimethylether.getFluid(1000))
+                .fluidInputs(EthyleneOxide.getFluid(1000))
+                .fluidOutputs(Dimethoxyethane.getFluid(1000))
+                .buildAndRegister();
+
+        CHEMICAL_RECIPES.recipeBuilder().duration(160).EUt(2000)
```
pablocnvs, line 27: Can you explain in more detail this step?

pablocnvs (commented):
Feels pretty good, there is only one step that i can't comprehend (maybe i should've known since that happens with my own charts) and, most importantly, it is imbalanced and not knowing the step you chose makes it difficult to balance.
Overall though, i don't see anything problematic

src/main/java/gregicadditions/recipes/chain/OrganometallicChains.java
Zalgo239: I think the issue here is the Dimethoxyethane, IRL Lithium Cyclopentadienide is usually handled as a THF solution, Dimethoxyethane is an alternative to THF. It's not consumed in the reaction itself, it's just meant to be a solvent.

pablocnvs: But if that is the case, why did you use cyclooctadiene?

Zalgo239: Yes it's meant to be cyclopentadienide, I'll fix it tomorrow

src/main/java/gregicadditions/recipes/RecipeHandler.java
```diff
@@ -156,6 +157,9 @@ public static void initChains() {
         ArcFurnaceOxidation.init();
         WetwareChain.init();
         OpticalChain.init();
+        OrganometallicChains.init();
+        QuantumDotsChain.init();
+        ExoticComponents.init();
```
DStrand1, line 162: Super minor, but `ExoticComponents.init()` should be called in `CircuitComponentRecipes.init()`. I left a code stub for you to add it right there :P

src/main/java/gregicadditions/recipes/categories/circuits/CircuitRecipes.java
```diff
@@ -984,12 +984,80 @@ private static void opticalCircuits() {
 
     private static void exoticCircuits() {
 
-        /* COMING SOON...
-
         for (FluidStack fluidStack : SOLDER_FLUIDS) {
             fluidStack.amount = Math.min(64000, fluidStack.amount * 4);
```
DStrand1, line 988:
```suggestion
            FluidStack fluidStackx4 = new FluidStack(fluidStack.getFluid(), Math.min(64000, fluidStack.amount * 4));
```
This is necessary for the soldering fluid bug to not return, where soldering fluids take insane amounts of fluid due to updating a pointer. Make sure to update the use below at line 998.

src/main/java/gregicadditions/recipes/chain/QuantumDotsChain.java
```diff
@@ -0,0 +1,51 @@
...
+
+import net.minecraftforge.fluids.FluidStack;
+
+import static gregicadditions.GAMaterials.*;
+import static gregicadditions.recipes.GARecipeMaps.*;
+import static gregtech.api.recipes.RecipeMaps.*;
+import static gregtech.api.unification.material.Materials.*;
+import static gregtech.api.unification.ore.OrePrefix.dust;
+
+public class QuantumDotsChain {
```
DStrand1, line 11: Another minor thing, but I think it makes more sense for this to be a method inside of `VariousChains` since it is only 5 recipes, unless you anticipate more recipes being added to this file later on

DStrand1 (changes requested):

src/main/java/gregicadditions/recipes/chain/QuantumDotsChain.java
Zalgo239: I thought about it too, the reason why I kept them in a separate class is that Quantum Dots have a bunch of applications I find interesting, they can be used as dyes, in LEDs, displays, solar cells, in the production of solar fuels, and have a bunch of other applications, I believe these could be of interest to us.

PR #529, opened 2021-06-11
Add linter tests to CI

DStrand1:
This PR adds a Linter GHA test to ensure good code quality going forward. It will not error on pre-existing, poorly formatted code, but it will stop any more bad code from going in. There are certainly a few settings with it that will need to be tweaked to be a bit less aggressive, but I think this is a good step towards better code overall.

Eventually I would like to add a gradle task to lint over your changes, so that contributors do not need to open a PR and hope for the best, and can instead lint their code before submitting. However this will probably require first linting over the entire codebase to make it consistent, which will be a large and disruptive PR that should happen when there are no large in-progress PRs.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -44,6 +44,7 @@ jobs:
   runServer:
     name: Runtime
     runs-on: ubuntu-latest
+    continue-on-error: true
 
     steps:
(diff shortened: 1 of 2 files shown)


Yefancy:
It's really necessary. Not only in Action, but also Gradle task.

PR #531, opened 2021-06-12
Add Mechanism for Properly Overriding GTCE Lang Entries

DStrand1:
Currently, if we set new translations for GTCE (or other mod/vanilla) items, the actual chosen translation is GUI dependent, and at times, seemingly random. This PR adds a mechanism for consistently overriding default lang implementations.

Adapted from the code, from a javadoc:

Examples of the methods:
- Normal Blocks and Items:
    - `setLocalization(EN_US, "tile.wire_coil.cupronickel.name", "funni coil");`
- GTCE MetaItems:
    - `setMetaItemName(RU_RU, MetaItems.CREDIT_COPPER.getMetaItem(), "модный кредит");`
- GTCE MetaItem Tooltips:
    - `setMetaItemTooltip(ZH_CN, MetaItems.CREDIT_COPPER.getMetaItem(), "最值钱!");`

These methods will work for all GTCE, vanilla, and other modded blocks, items, and any other potential translation keys we could ever want to override.


DStrand1:
For now, I believe we can merge without linter passing, so we may want to take some of its reviews with a grain of salt for now, as it is quite aggressive with some of its code cleanliness enforcement

PR #533, opened 2021-06-13
Add mechanism to add more slots to GTCE RecipeMaps

DStrand1:
This PR adds a (kind of terrible, but functional) way of adding new slots to GTCE RecipeMaps. Currently, I have it adding:
- 1 item output to Distillery
- 1 fluid output to Chemical Bath
- +1 item output to Chemical Reactor

image
image
image

### DIFF
--- a/src/main/java/gregicadditions/GAEnums.java
+++ b/src/main/java/gregicadditions/GAEnums.java
@@ -1,5 +1,6 @@
 package gregicadditions;
 
+import gregtech.api.recipes.RecipeMap;
 import gregtech.api.unification.Element;
 import gregtech.api.unification.material.MaterialIconType;
 import gregtech.api.unification.material.Materials;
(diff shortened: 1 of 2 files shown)


src/main/java/gregicadditions/GAEnums.java
```diff
@@ -146,8 +149,35 @@ public static void preInit() {
...
+     * Sets a RecipeMap to have a different number of slots for inputs or outputs.
+     * USE THIS SPARINGLY!
+     *
+     * @param map      The RecipeMap to set the field of.
+     * @param slotType The slot (maxInputs, maxOutputs, etc.) to set.
+     * @param value    The new value of slotType
+     *
+     * @throws Exception If failed.
+     */
+    public static void addSlotsToGTCEMaps(
```
hjae78, line 164: I don't know whether this should be in this file really, maybe better off in a utility file?

DStrand1: I stuck it here because this is where most of our other reflection is located. I agree that with this the name for this file doesn't make much sense, but I think it fits well with the other things in this file. Maybe we should instead rename the file?

hjae78: Ye that might be a better idea

PR #537, opened 2021-06-17
[TOOL] Progressive Prospector

Yefancy:
### Changes as follow:

1. Server-Friendly: Progressive loading to avoid server laggy. 
2. Configuration: Add a config to set the scanning speed.
3. UI: Adjust the minimum size of the window so that the List could display more information.
4. Creative: Zero cost in creative mode.

### Screenshots
progressive


htmlcsjs:
Why not save darkmode as nbt?

Yefancy:
> Why not save darkmode as nbt?

The main reason is that this feature was added temporarily. I don't think persistence is important, after all, it's only designed to help the player recognize colors that are hard to see.

htmlcsjs:
Yea l prefer darkmode, so it would be nice if it didn't reset every time I use jt

PR #550, opened 2021-07-01
Update Electric Implosion Compressor

ALongStringOfNumbers:
Update the Electric Implosion Compressor after the release of the newest GTCE version, which transitioned the Implosion Compressor over to the new property system.

The try/catch was removed because the explosive was transitioned to an itemstack, and the TNT would be the correct itemstack if a custom explosive was not specified.


DStrand1 (approved):

TechLord22 (approved):

hjae78:
Speaking of Electric Implosion bugs shouldn't we take a stab at solving the JEI itemstack issue? I just quickly had a look and noticed that we're modifying the stack sizes of some items in this file. We should try to see what happens if we instead make new copies of the items with different sizes. I don't know if this is appropriate in this PR or not

ALongStringOfNumbers:
With the most recent change, I no longer saw the JEI itemstack issue where items would display with a stacksize > 1 when looking at recipes in JEI from a stack of gems with size > 1.

PR #556, opened 2021-07-04
[WAFER OVERHAUL] CVD Unit and basic wafer chains

Puzzlebark:
This PR implements the structure and recipe map for the CVD unit, and implements processing chains for a few basic wafers (ILC, CPU, RAM).

The current recipes, multiblock structure and dopant mechanics are not final.


src/main/java/gregicadditions/recipes/chain/WaferChain.java
```diff
@@ -1,20 +1,51 @@
...
+                .EUt(120)
+                .buildAndRegister();
+
+        ASSEMBLER_RECIPES.recipeBuilder()
+                .inputs(OreDictUnifier.get(plateDense, Aluminium))
+                .notConsumable(craftingLens, Silver)
+                .outputs(LITHOGRAPHY_MASK_RAM.getStackForm())
+                .duration(900)
+                .EUt(120)
+                .buildAndRegister();
```
hjae78, line 46: I think these could be engraver recipes

PR #557, opened 2021-07-05
Void miner stable work fix

hhheeexxx:
Small fix for void miner.
problem:
The void miner was constantly cooling down due to asymmetric heating and cooling with the same amount of liquids. Without explicit type casting, an inaccuracy occurred: the cooling was 1 degree more than the heating. And 1 / 1.02 not equal 1 * 0.98.


hjae78:
This is the point of the void miner, to make it more difficult to automate. I have added redstone output to void miners relative to their current temp to make this possible. It is also possible to read the temperature using OpenComputers.

hhheeexxx:
There is a problem that the temperature drops, and the amount of liquid consumed does not fall, as a result, more and more liquid is needed to maintain the miner at a constant temperature. As a result, it will be necessary to periodically overheat the system in order to reset the liquid meter. If using open computers, this can be automated by overheating the system on a timer every few days. But if you do it without open computers, then you need to output the amount of consumed liquid through the redstone, or make a strict linear relationship between the temperature and the consumed liquid, or add a mod for a long timer, or some solution that I have not listed. I would prefer a linear relationship or redstone output of the amount of fluid consumed. What do you think?

hjae78:
Yes I agree that this isn't ideal. What about adding an upper limit on the fluid consumed?

TechLord22 (approved):

PR #561, opened 2021-07-09
Advanced Fusion Fixes

pablocnvs:
Quickfix for initial EUt and adds tiers to the JEI preview, helpful for the freedom wrench


ALongStringOfNumbers:
Hm, is the eu to start property used for these reactors the same that is used for the regular fusion reactors, or is it different?

pablocnvs:
I think it is a different MTE type with a different recipemap, but it pretty much functions the same way. The pr fixes the fact that the eut to start was much higher than what you could get with the coil and hatch of that tier.

hjae78 (approved):
Looks good to me!

PR #582, opened 2021-08-10
remove the cheaper assembler component recipes and config

BraggestSage833:
What: Removes the cheaper assembler component recipes and corresponding config

Outcome: Resolves the conflicts with the component recipes added by GTCE

Possible compatibility issue: May possibly break players setups

### DIFF
--- a/src/main/java/gregicadditions/GAConfig.java
+++ b/src/main/java/gregicadditions/GAConfig.java
@@ -355,11 +355,6 @@ public class GAConfig {
         @Config.RequiresMcRestart
         public boolean assemblyLineMakeCheaperComponents = true;
 
-        @Config.Comment("Assembler can make components (Motor, Pump, Conveyor, Piston, etc)")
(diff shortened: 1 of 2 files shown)


ALongStringOfNumbers (changes requested):
This PR also needs to adjust the assembler pump recipes, as Gregicality adjusts the crafting table recipes to be GT5U recipes, therefore the assembler recipes need to be adjusted to match. See this PR https://github.com/Shadows-of-Fire/Shadows-of-Greg/pull/154 that I did for SoG doing the same thing.

In addition, the removal of Field Generator assembler recipes from GTCE needs to stop, so that the recipes actually exist.

For a bunch of these components (SoG did all except for the pumps) Gregicality is removing the GTCE crafting recipe and then adding back the exact same recipe, which should be something else that is dumped in this PR

DStrand1 (approved):

PR #588, opened 2021-08-15
Start second degree materials

kumquat-ir:
Everything before the first simpleMaterial


src/main/java/gregicadditions/materials/GASecondDegreeMaterials.java
```diff
@@ -1,8 +1,421 @@
...
 public class GASecondDegreeMaterials {
 
+    /*
+     * IDs 11500-14499
+     */
+
     public static void register() {
+        PalladiumAmmonia = new Material.Builder(11500, "palladium_enriched_ammonia")
+                .fluid()
+                .color(Platinum.getMaterialRGB())
```
TechLord22, line 34: Call the colorAverage() method instead.

PR #602, opened 2021-09-30
Fix Maintenance Total Time Not Resetting

TechLord22:
This PR fixes the total time on for multiblock maintenance not being reset after a problem occurs. This would previously cause the multiblock to constantly have problems after the first one arises, and has now been fixed.

This fix is already present in CEu, and was taken from there.


pablocnvs (approved):
Go on, looks good

BraggestSage833 (approved):

PR #625, opened 2022-02-04
Nuclear reprocessing changes

BraggestSage833:
What:
Buffs spent fuel repossessing
How:
I lowered the EU/t and duration of fuel reprocessing recipes as they are currently the largest bottleneck of nuclear.
increased the waste created when turning spent fuel nitrate solution(into spent fuel nitride by one. made the fluids and ferrite dust that would normally be looped through red oil not consumed to further increase parallelization of these recipes.


pablocnvs (approved):
Fine to me

TechLord22 (changes requested):
These changes look good, except that all of the recipes changed need to use `30` EU/t instead of `32` for cable loss purposes.

TechLord22 (approved):

